PR #36, opened 2014-12-31
Add upper-case and lower-case

orangejenny:
FB151803


core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -821,6 +825,23 @@ public static String substring (Object o1, Object o2, Object o3) {
...
+
+    /**
+     * Perform toUpperCase or toLowerCase on given object.
+     * @param o
+     * @param toUpper
+     * @return
+     */
+    private String normalizeCase (Object o, boolean toUpper) {
+        String s = toString(o);
+        if (s == null) {
```
ctsims: It should be impossible for s to be null after a call to toString. I can update the docs for that method.

orangejenny: fixed

PR #38, opened 2015-01-05
Function bevy

orangejenny:
Add a bunch of math & string functions that 
a) seem useful
and
b) are easy to implement


core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -247,6 +247,16 @@ public Object eval (DataInstance model, EvaluationContext evalContext) {
                 return normalizeCase(argVals[0], true);
             } else if (name.equals("lower-case") && args.length == 1) {
                 return normalizeCase(argVals[0], false);
+            } else if (name.equals("contains") && args.length == 2) {
```
orangejenny: fixed

orangejenny: Missed a contains and missed that Math.round doesn't exist in 1.3.

Fixed for real now.

PR #41, opened 2015-01-09
Fix for min/max not working with paths and argument lists

ctsims:
Removed HG requirement from build process.

Removed crypto tests, since code is no longer included in builds

Updated XPath Eval code to include potential for path driven tests again

Added subreference tests to treereference class tests

added tests for nodeset v. arg list polymorphism

updated treereference docs

Fixed bug with min/max/sum functions taking in argument lists that are paths

### DIFF
--- a/core/test/org/javarosa/xpath/test/XPathEvalTest.java
+++ b/core/test/org/javarosa/xpath/test/XPathEvalTest.java
@@ -26,8 +26,12 @@ import java.util.Vector;
 
 import org.javarosa.core.model.condition.EvaluationContext;
 import org.javarosa.core.model.condition.IFunctionHandler;
+import org.javarosa.core.model.data.IAnswerData;
+import org.javarosa.core.model.data.IntegerData;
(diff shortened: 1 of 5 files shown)


core/test/org/javarosa/xpath/test/XPathEvalTest.java
```diff
@@ -547,57 +563,23 @@ private void addDataRef (FormInstance dm, String ref, IAnswerData data) {
         addNodeRef(dm, ref, true);
         
         if (data != null) {
-            dm.updateDataValue(new XPathReference(ref), data);
+            dm.resolveReference(new XPathReference(ref)).setValue(data);
         }
     }
     
     private void addNodeRef (FormInstance dm, String ref, boolean terminal) {
```
orangejenny: terminal isn't being used anymore

ctsims: thanks, fixed!

PR #42, opened 2015-01-22
Audio logs

orangejenny:
Produces logs like `media> start something.mp3`

Left the filename extension so it's easier to distinguish between audio and video.

### DIFF
--- a/j2me/form-entry/src/org/javarosa/formmanager/api/JrFormEntryController.java
+++ b/j2me/form-entry/src/org/javarosa/formmanager/api/JrFormEntryController.java
@@ -3,22 +3,17 @@
  */
 package org.javarosa.formmanager.api;
 
-import java.io.IOException;
 import java.util.Vector;
 
 import javax.microedition.lcdui.Canvas;
-import javax.microedition.media.Manager;
 import javax.microedition.media.MediaException;
 import javax.microedition.media.Player;
+import javax.microedition.media.PlayerListener;
 
-import org.javarosa.core.api.State;
-import org.javarosa.core.data.IDataPointer;
 import org.javarosa.core.model.FormIndex;
 import org.javarosa.core.model.SelectChoice;
 import org.javarosa.core.model.utils.DateUtils;
-import org.javarosa.core.reference.InvalidReferenceException;
 import org.javarosa.core.reference.Reference;
-import org.javarosa.core.reference.ReferenceManager;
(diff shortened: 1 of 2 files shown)


j2me/form-entry/src/org/javarosa/formmanager/api/JrFormEntryController.java
```diff
@@ -294,6 +276,8 @@ public void attachVideoPlayer(Player player) {
         if(this.player != null) {
             detachVideoPlayer(player);
         }
+        player.removePlayerListener(mediaLogger);
+        player.addPlayerListener(mediaLogger);
```
ctsims: ? Probably worth a comment at least to specify why we need to remove then add?

j2me/form-entry/src/org/javarosa/utilities/media/MediaUtils.java
```diff
@@ -136,13 +140,47 @@ private static String getFileFormat(String fpath){
...
+                action = "stop";
+            }
+            if (!"".equals(action)) {
+                try {
+                    Vector<String> pieces = DateUtils.split(uri, "/", false);
+                    uri = pieces.lastElement();
+                } catch (Exception e) {
+                    // just use the full URI
+                }
+                Logger.log("media", action + " " + uri);
```
ctsims: As much as I hate magic strings, it might make the data team's job a bit easier if we delimit this with something that is unlikely to exist on either side (It's possible and not super uncommon for the URI's to have spaces in them, I think)

j2me/form-entry/src/org/javarosa/formmanager/api/JrFormEntryController.java
```diff
@@ -53,6 +45,8 @@
     
     private static int POUND_KEYCODE = Canvas.KEY_POUND;
     
+    private PlayerListener mediaLogger;
```
ctsims: mMediaLogger?

ctsims:
good to go after comments

j2me/form-entry/src/org/javarosa/utilities/media/MediaUtils.java
orangejenny: Good point. Looks like colons are illegal on windows & mac, switching format to `media> start: something.mp3`

orangejenny:
@ctsims addressed comments, should be all set

PR #43, opened 2015-02-11
Fixture predicate bug fix

ctsims:
Bugfix for serious predicate evaluation issue where in some circumstances predicates in external instances would merge subsequent predicates rather than perform a union/filter

### DIFF
--- a/core/src/org/javarosa/core/util/DataUtil.java
+++ b/core/src/org/javarosa/core/util/DataUtil.java
@@ -15,7 +15,7 @@ public class DataUtil {
     static final  int high = 400;
     static Integer[] iarray;
     
-    static UnionLambda unionLambda = new UnionLambda();;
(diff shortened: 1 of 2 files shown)


core/src/org/javarosa/core/util/DataUtil.java
```diff
@@ -15,6 +15,12 @@
     static final  int high = 400;
     static Integer[] iarray;
     
+    static UnionLambda unionLambda;
+    
+    { 
+        unionLambda = new UnionLambda();
```
orangejenny: does having braces around this do anything?

ctsims: Oh, sorry, this is a static initializer (it's actually unrelated to the declaration above), but it occurs to me that it should actually just be on the static declration. Will fix.

PR #45, opened 2015-02-19
Markup 2

wpride:


wpride:
@ctsims encoding updated

core/src/org/javarosa/core/services/locale/LocalizationUtils.java
```diff
@@ -72,10 +72,12 @@ private static void parseAndAdd(OrderedHashtable locale, String line, int curlin
...
             //trim whitespace.
             line = line.trim();
+
+            int i = 0;
             
-            //clear comments
-            while(line.indexOf("#") != -1) {
-                line = line.substring(0, line.indexOf("#"));
+            //clear comments except if they have backslash before them (markdown '#'s)
+            while((i = line.indexOf("#")) != -1 && !(line.charAt(i-1) == '\\')) {
```
ctsims: if the "#" is at 0, this will crash.

We should add tests for parsing lines to the test suite for javarosa in general to catch everything

core/test/org/javarosa/core/util/test/LocalizationTests.java
```diff
@@ -0,0 +1,116 @@
...
+
+        assertTrue("Parsed newlines correctly", result1.equals("1. One \n2. Two \n3. Three"));
+        assertTrue("Parsed newlines correctly", result2.equals("# Start"));
+    }
+
+    public void testParseAndAdd(){
+
+        OrderedHashtable<String, String> testTable = new OrderedHashtable<String, String>();
+
+        //LocalizationUtils.parseAndAdd(testTable,"string.0=#this line should be all comment", 0);
```
ctsims: Why did this line get removed? It seems like the we should have a test that starts with "#" just to make sure it doesn't crash anything.

like

"#comment goes here"

wpride:
At the moment having value=#comment does crash the engine with error..Invalid line (#0) read: 'string.0='. No value follows the '='.

PR #46, opened 2015-02-27
Output dep cycle parse error

phillipm:
Fixing bug where when <output> structures in labels are malformed (example below), the xpath parser throws a message which bubbles up but makes no sense to the end user,

"Dependency cycle in <output>s; recursion limit exceeded!!"

This message should instead simply specify something like:

"Malformed expression: TEXT_OF_EXPRESSION_HERE"

Example:
<output ref="instance('casedb')/casedb/case[@case_id=current()/../../@id)]/case_name"/>

### DIFF
--- a/core/src/org/javarosa/core/model/FormDef.java
+++ b/core/src/org/javarosa/core/model/FormDef.java
@@ -1002,15 +1002,36 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
     public String fillTemplateString(String template, TreeReference contextRef) {
         return fillTemplateString(template, contextRef, new Hashtable());
     }
-    
+
+    /**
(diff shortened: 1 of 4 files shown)


core/src/org/javarosa/core/model/FormDef.java
```diff
@@ -1002,15 +1002,36 @@ public boolean realTime() {
...
+     * Performs substitutions on place-holder template from form text by
+     * evaluating args in template using the current context.
+     *
+     * @param template String
+     * @param contextRef TreeReference
+     * @param variables Hashtable<String, ?>
+     *
+     * @return String with the all args in the template filled with appropriate
+     * context values.
+     */
```
snopoke: Nice method comment :+1:

snopoke:
No unit tests :disappointed:

PR #47, opened 2015-03-03
Xform engine recognize intents

phillipm:
Fixing [156946](http://manage.dimagi.com/default.asp?156946), where the form_validator tool was warning about "Unrecognized question type" on extensions to the xform specification, such as "intents"

This also includes a bunch of clean up and code style changes.

Examples:
The following exhibits the warnings we want to suppress (where the inputted xml is the External Callout form from corpora on HQ):
$ java -jar form_translate.jar validate external_callout_intent.xml

To suppress the warnings we extend the spec, via command-line flags, with place-holder information:
$ java -jar form_translate.jar -E http://opendatakit.org/xforms=intent -E http://www.w3.org/2002/xforms=extra -S http://opendatakit.org/xforms -I http://opendatakit.org/xforms validate ~/dimagi/xforms/external_callout_intent.xml
where:
 -E says we are expecting "intent" tags at the namespace "http://opendatakit.org/xforms"
 -S says suppress warnings that come from parsing expected elements in the "http://opendatakit.org/xforms" namespace
 -I says continue parsing inner elements of expected tags from the "http://opendatakit.org/xforms" namespace


phillipm:
Cleaned up the PR and reposted here:
https://github.com/dimagi/javarosa/pull/68

PR #52, opened 2015-03-09
Touchforms access fix

wpride:


czue:
noice.

ctsims:
@orangejenny Hey, can you double check real quick when you get a change whether there are any negative consequences to making this method public/whether that was intended?

PR #53, opened 2015-03-09
Adding core tests build

phillipm:
Round 2, now without extraneous commits from other branches :)
Unit tests can now be built/run from core build file via 'ant test'

Notes:
 - Commit c748cb9 removes 'clean' as a dep for core project compilation. Will this break any automated builds we have in places like jenkins?
 - The only changes to j2me/buildfiles/build.xml are in commit  418b333 and are purely whitespace and retabbing

### DIFF
--- a/core/build.xml
+++ b/core/build.xml
@@ -52,7 +52,7 @@
         <delete dir="${build}"/>
     </target>
 
-    <target name="compile" depends="clean, extract-libs, init" description="compile the source">
+    <target name="compile" depends="extract-libs, init" description="compile the source">
         <!-- To get rid of
              "bootstrap class path not set in conjunction with -source 1.5" warning
              add bootclasspath="path/to/java/instance/rt.jar" attribute to javac tag -->
--- a/core/build.xml
+++ b/core/build.xml
@@ -1,14 +1,42 @@
 <?xml version="1.0"?>
 <project default="package">
+    <!-- ///////////////////////////////////////////// -->
+    <!-- Setup                                         -->
+    <!-- ///////////////////////////////////////////// -->
+    <!-- directory locations -->
     <property name="build" location="bin"/>
     <property name="dist" location="dist"/>
     <property name="src" location="src"/>
-
-
-    <property name="name" value="javarosa-libraries"/>
-    <property name="jar" value="${name}.jar"/>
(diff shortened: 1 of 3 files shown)


core/build.xml
```diff
@@ -1,40 +1,180 @@
...
-	     are using the version of the archive compatible with the version you wish to build) -->
-	<target name="extract-libs" if="extract-libs-fresh">
-		<unzip src="${javarosa-deps}" dest="${dir.root}" overwrite="true" />
-	</target>
-	
+    <!-- ///////////////////////////////////////////// -->
+    <!-- Setup                                         -->
+    <!-- ///////////////////////////////////////////// -->
+    <!-- directory locations -->
+    <property name="build" location="bin"/>
```
ctsims: General best practice is to configure this kind of stuff in a build.properties file that you include adjacent to this file, and then also load a "local.properties" file that is in gitignore so we can override stuff

core/build.xml
```diff
@@ -1,40 +1,180 @@
...
+        <include name="**/resources/*.x*ml/"/>
+    </fileset>
+
+
+    <!-- ///////////////////////////////////////////// -->
+    <!-- Build targets                                 -->
+    <!-- ///////////////////////////////////////////// -->
+
+    <target name="init" description="set classpath and make needed directories">
+        <mkdir dir="lib/"/>
```
ctsims: This is a temporary ${dir.lib} or should this also be that named property?

core/build.xml
```diff
@@ -1,40 +1,162 @@
...
-	<!-- for use primarily by the build server to extract the required 3rd-party libraries into the
-	     javarosa source tree. to use this in your own build, set the 'extract-libs-fresh' property,
-	     then also set the 'javarosa-deps' property to the path of the library archive (make sure you
-	     are using the version of the archive compatible with the version you wish to build) -->
-	<target name="extract-libs" if="extract-libs-fresh">
-		<unzip src="${javarosa-deps}" dest="${dir.root}" overwrite="true" />
-	</target>
-	
+
+    <property file="build.properties"/>
```
ctsims: can you load local.properties as well here, in case people want to override without getting their current working set dirty?

PR #54, opened 2015-03-11
Basic tests for answerQuestion method of FormEntryController

phillipm:
This PR is a basic test for answering simple questions with and without constraints.

FormEntryController.answerQuestion has some strange logic in it that has to do with answering questions that are complex. For background complex questions have copy tags in them that need xml subtree substitution to occur during evaluation.

I wanted to test that logic, gave up after not being able to produce a valid xform as input.


millerdev:
Tests! Sweet!

snopoke:
:star2:

ctsims:
@phillipm Are these ready for merging? Don't see the flag, but look good to me.

phillipm:
@ctsims yup, I created this PR before knowing about labels. Should be ready to review/merge.

PR #58, opened 2015-03-20
Truncate geopoint alt & acc data to nearest millimeter

phillipm:
The phone is recording small doubles for the altitude and accuracy fields sometimes.

This PR rounds up those values to two decimal points, since we really don't care about accuracy beyond that for these fields.

[Bug 161590](http://manage.dimagi.com/default.asp?161590)


core/src/org/javarosa/core/model/data/GeoPointData.java
```diff
@@ -132,4 +148,21 @@ public GeoPointData cast(UncastData data) throws IllegalArgumentException {
...
+
+    /**
+     * Truncate double to have the given number of decimals.
+     * @param x double to be truncated
+     * @param numberofDecimals number of decimals that should present in result
+     */
+    private static double truncateDecimal(double x, int numberofDecimals) {
+        // via:
+        // https://stackoverflow.com/questions/7747469/how-can-i-truncate-a-double-to-only-two-decimal-places-in-java/21468258#21468258
+        BigDecimal decimal;
```
ctsims: I don't think BigDecimal's are Java 1.3 compliant, btw, so we probably need to find a J2ME compatible way to do this truncation/rounding.

PR #61, opened 2015-03-28
check function arity

wpride:


wpride:
Not ready to be merged - just wanted to open to comment before bothering to go through all of the functions

core/src/org/javarosa/xpath/XPathArityException.java
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Created by wpride1 on 3/28/15.
+ */
+public class XPathArityException extends XPathException {
+    public XPathArityException() {
+
+    }
+
+    public XPathArityException(String s, int required, int actual) {
+        super("cannot handle " + s + " had " + actual + " arguments but required " + required + ".");
```
ctsims: In order to make sure this is localize, it would likely be helpful to keep these parameters (function name, expected, provided), so we can catch the typed exception and throw a message with those details that's localized in the end platform (in addition to having this message for technical/fallback info)

phillipm:
Thought this was ready for review, so I cleaned it up so that we can empty out the PR queue and restyle the whole code base.

wpride:
nice, thanks phillip!

PR #67, opened 2015-04-03
template path traversal isn't robust enough to handle hetergeneous xml structures

phillipm:
Built a corollary function to getTemplatePath called hasTemplatePath. 

hasTemplatePath traverses the complete document tree, not just the first branch, in search of a path that satisfies the reference in question, returning a boolean answer, instead of an (unused) element object.

[fixes this ticket](http://manage.dimagi.com/default.asp?162758)


core/src/org/javarosa/core/model/instance/DataInstance.java
```diff
@@ -168,7 +168,11 @@ public Vector explodeReference(TreeReference ref) {
 
     public T getTemplate(TreeReference ref) {
         T node = getTemplatePath(ref);
-        return (node == null ? null : ((node.isRepeatable() || node.isAttribute()) ? node : null));
+
+        if (node == null || !(node.isRepeatable() || node.isAttribute())) {
+            return null;
```
phillipm: Boolean refactor (which can always use a second pair of eyes)

core/src/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -214,59 +214,15 @@ public XPathNodeset eval(DataInstance m, EvaluationContext ec) {
...
-        if (ref.isAbsolute()) {
-            if (cachedTemplate == null) {
-                cachedTemplate = m.getTemplatePath(ref);
-            }
-            if (cachedTemplate == null) {
-                return XPathNodeset.ConstructInvalidPathNodeset(ref.toString(), genericRef.toString());
-            }
+        // Error out if there doesn't exist a (template) path along the
+        // reference starting at the main DataInstance.
+        if (ref.isAbsolute() && !m.hasTemplatePath(ref)) {
```
phillipm: Removed strange unused cacheTemplate var.
Please check that my new logic mirrors the old logic

core/src/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -214,59 +214,15 @@ public XPathNodeset eval(DataInstance m, EvaluationContext ec) {
...
+            return XPathNodeset.constructInvalidPathNodeset(ref.toString(), genericRef.toString());
```
phillipm: Boolean refactor (which can always use a second pair of eyes)

core/src/org/javarosa/core/model/instance/DataInstance.java
```diff
@@ -168,7 +168,11 @@ public Vector explodeReference(TreeReference ref) {
 
     public T getTemplate(TreeReference ref) {
         T node = getTemplatePath(ref);
-        return (node == null ? null : ((node.isRepeatable() || node.isAttribute()) ? node : null));
+
+        if (node == null || !(node.isRepeatable() || node.isAttribute())) {
+            return null;
+        }
+        return node;
```
phillipm: Boolean refactor (which can always use a second pair of eyes)

ctsims: looks good

core/src/org/javarosa/xpath/expr/XPathPathExpr.java
ctsims: (Talked on skype, but for posterity): We do still want to be caching the template check still, just in the new boolean form.

core/test/org/javarosa/xpath/expr/test/XPathPathExprTest.java
```diff
@@ -91,39 +92,22 @@ public void run(TestCase tc) {
...
+            fail("Form at " + formPath + " has an invalid structure.");
         }
 
-        // Used to reproduce bug where locations can't handle heterogeneous template paths
-        // The following test will pass when this is fixed.
-        testEval("/data/places/country[1]/state[0]/name", instance, null, "Utah");
+        // Used to reproduce bug where locations can't handle heterogeneous template paths.
+        // This bug has been fixed and the following test now passes. Expected value should
+        // be Utah, but multiplicity selectors aren't implemented...
+        testEval("/data/places/country[1]/state[0]", instance, null, "");
```
ctsims: So I wouldn't trust these tests to be telling you anything until we have the syntactic sugar for this enabled.

Easiest replication would be to test with a predicate selector. IE:

<data>
   <model id='one'>
        <subvalue>
   </model>
   <model id='two'>
        <subvalue>
....

and use the [@id ='two'] predicate, instead of [1]

PR #68, opened 2015-04-07
Xform engine recognize intents (round 2)

phillipm:
Fixing [156946](http://manage.dimagi.com/default.asp?156946), where the form_validator tool was warning about "Unrecognized question type" on extensions to the xform specification, such as "intents"

This also includes a bunch of clean up and code style changes.

Examples:
The following exhibits the warnings we want to suppress (where the inputted xml is the External Callout form from corpora on HQ):

$ java -jar form_translate.jar validate ../../../core/resources/intent_example.xml

To suppress the warnings we extend the spec, via command-line flags, with place-holder information:
$ java -jar form_translate.jar -E http://opendatakit.org/xforms=intent -E http://www.w3.org/2002/xforms=extra -S http://opendatakit.org/xforms -I http://opendatakit.org/xforms validate ../../../core/resources/intent_example.xml

where:
 -E says we are expecting "intent" tags at the namespace "http://opendatakit.org/xforms"
 -S says suppress warnings that come from parsing expected elements in the "http://opendatakit.org/xforms" namespace
 -I says continue parsing inner elements of expected tags from the "http://opendatakit.org/xforms" namespace


core/src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -503,18 +615,24 @@ private void parseElement (Element e, Object parent, Hashtable<String, IElementH
...
         } else {
-            if (!suppressWarning.contains(name)) {
-                reporter.warning(XFormParserReporter.TYPE_UNKNOWN_MARKUP,
-                        "Unrecognized element [" + name    + "]. Ignoring and processing children...",
-                        getVagueLocation(e));
-            }
-            for (int i = 0; i < e.getChildCount(); i++) {
-                if (e.getType(i) == Element.ELEMENT) {
-                    parseElement(e.getElement(i), parent, handlers);
+            if (inSpecExtension(namespace, name)) {
```
phillipm: Entry point into the bug fix

core/resources/intent_example.xml
```diff
@@ -0,0 +1,83 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
```
ctsims: Is this doc used anywhere? 

Might be kind of hard to actually test the inputs/outputs automatically, just checking whether we had a clear meaning for this folder.

phillipm: I added this file so that one can easily (manually) test/see what the PR is achieving (see example usage in PR description). I should probably just write real tests, but when I started this PR the unit testing framework wasn't up and running again.

PR #71, opened 2015-04-10
Cursory clean-up of TreeReference code

phillipm:
In prep for working on some reference related bugs I'm reading through code, cleaning up comments and adding tests

I've noted changes that are worthy of a 2nd look

### DIFF
--- a/core/src/org/javarosa/core/model/instance/TreeReference.java
+++ b/core/src/org/javarosa/core/model/instance/TreeReference.java
@@ -66,10 +66,12 @@ public class TreeReference implements Externalizable {
     //TODO: Roll these into RefLevel? Or more likely, take absolute
     //ref out of refLevel
     public static final int CONTEXT_ABSOLUTE = 0;
+    // context is inherited since the path is relative
     public static final int CONTEXT_INHERITED = 1;
+    // use the original context instead of current context, used by the
+    // current() command.
     public static final int CONTEXT_ORIGINAL = 2;
     public static final int CONTEXT_INSTANCE = 4;
-
 
     public static final int REF_ABSOLUTE = -1;
 
@@ -148,6 +150,7 @@ public class TreeReference implements Externalizable {
     /**
      * How many reference levels are present? Compute this value on demand and
      * cache it.
+     *
      * @return the number of reference levels
      */
     public int size() {
@@ -173,11 +176,11 @@ public class TreeReference implements Externalizable {
     }
 
     /**
-     * Store a copy of the reference level at level 'key'. 
+     * Store a copy of the reference level at level 'key'.
      *
      * @param key reference level at which to attach predicate vector argument.
      * @param xpe vector of xpath expressions representing predicates to attach
-     * to a reference level.
+     *            to a reference level.
      */
     public void addPredicate(int key, Vector<XPathExpression> xpe) {
         hashCode = -1;
@@ -258,7 +261,7 @@ public class TreeReference implements Externalizable {
         return false;
     }
 
-    /** 
+    /**
      * return a copy of the ref
      */
     public TreeReference clone() {
@@ -278,7 +281,7 @@ public class TreeReference implements Externalizable {
         return newRef;
     }
 
-    /** 
+    /**
      * Return a copy of the TreeReference that doesn't include any of the
      * TreeReferenceLevels. Useful when we are just going to overwrite the
      * levels with new data anyways.
@@ -360,18 +363,19 @@ public class TreeReference implements Externalizable {
     }
 
 
-    /** 
-    * Similar to parent(), but assumes contextRef refers to a singular,
-    * existing node in the model.  This means we can do '/a/b/c + ../../d/e/f =
-    * /a/d/e/f', which we couldn't do in parent().
-    * Returns null if context ref is not absolute, or we parent up past the root
-    * node.
-    * NOTE: this function still works even when contextRef contains
-    * INDEX_UNBOUND multiplicites...  conditions depend on this behavior, even
-    * though it's slightly icky
-    * @param contextRef absolute reference
-    * @return
-    */
+    /**
+     * Similar to parent(), but assumes contextRef refers to a singular,
+     * existing node in the model.  This means we can do '/a/b/c + ../../d/e/f =
+     * /a/d/e/f', which we couldn't do in parent().
+     * Returns null if context ref is not absolute, or we parent up past the root
+     * node.
+     * NOTE: this function still works even when contextRef contains
+     * INDEX_UNBOUND multiplicites...  conditions depend on this behavior, even
+     * though it's slightly icky
+     *
+     * @param contextRef absolute reference
+     * @return
+     */
     public TreeReference anchor(TreeReference contextRef) {
         // TODO: Technically we should possibly be modifying context stuff here
         // instead of in the xpath stuff;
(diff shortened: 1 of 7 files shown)


core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -629,31 +665,31 @@ public int getContext() {
...
-        ret.instanceName = this.instanceName;
-        ret.data = new Vector<TreeReferenceLevel>();
+        TreeReference subRef = new TreeReference();
+        subRef.refLevel = this.refLevel;
+        subRef.contextType = this.contextType;
+        subRef.instanceName = this.instanceName;
+
         for (int i = 0; i <= level; ++i) {
-            ret.data.addElement(this.data.elementAt(i));
+            subRef.add(this.data.elementAt(i).shallowCopy());
```
phillipm: We want to add this call to shallowCopy, right?

phillipm: Delegating through the 'add' method, instead of data.addElement, since it performs potentially crucial internal manipulations (hashCode = -1, size = -1, etc). 
Was there a (performance) reason that this wasn't being done before?

core/test/org/javarosa/core/model/test/TreeReferenceTest.java
```diff
@@ -152,139 +167,283 @@ public void doTest(int i) {
...
+        // setup expected result /a[1]/a[5] reference
+        TreeReference expectedAs = a.extendRef("a", 5);
+        System.out.println("setup path that looks like /a[1]/a[5]");
+        System.out.println(expectedAs.toString());
+
+        // ('../../a[5]').contextualize('/a[1]/a[2]/a[3]') ==> /a[1]/a[5]
+        // XXX: ^ that is right the expected behavior, right?
+        //      not /a[1]/a[2], which is what we get
+        contextualizeEval = a5.contextualize(aaa);
+        assertTrue("Got " + contextualizeEval.toString() + " and expected /a[1]/a[5]" +
```
phillipm: Potential bug that has to do with contextualize name matching logic

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -525,8 +649,12 @@ public String toString(boolean includePredicates) {
                         sb.append("[@juncture]");
                         break;
                     default:
-                        if ((i > 0 || mult != 0) && mult != -4)
+                        // Don't show a multiplicity selector if we are
+                        // selecting the 1st element, since this is the default
+                        // and showing brackets might confuse the user.
+                        if ((i > 0 || mult != 0) && mult != -4) {
```
phillipm: Is this comment correct?

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -337,18 +442,19 @@ else if (!this.getInstanceName().equals(contextRef.getInstanceName())) {
...
             }
 
             if (contextRef.getName(i).equals(newRef.getName(i))) {
-                //We can't actually merge nodes if the newRef has predicates or filters
-                //on this expression, since those reset any existing resolutions which
-                //may have been done.
+                // We can't actually merge nodes if the newRef has predicates
+                // or filters on this expression, since those reset any
+                // existing resolutions which may have been done.
                 if (newRef.getPredicate(i) == null) {
```
phillipm: Place where I think the failing contextualize test identifies a bug

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -279,57 +369,72 @@ public TreeReference parent(TreeReference parentRef) {
...
+            // copy level data from this ref to the anchor ref
+            for (int i = 0; i < size(); i++) {
+                newRef.add(this.data.elementAt(i).shallowCopy());
             }
+            return newRef;
         }
     }
 
-    //TODO: merge anchor() and parent()
-
```
phillipm: anchor and parent have pretty difference behavior when handling relative reference arguments, don't think they should eventually be merged

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -252,25 +335,32 @@ public TreeReference getParentRef() {
...
+     * @param baseRef an absolute reference or a relative reference with only
+     * '../'s
+     * @return a join of this reference with the base reference argument.
+     * Returns a clone of this reference if it is absolute, and null if this
+     * reference has '../'s but baseRef argument a non-empty relative reference.
+     */
+    public TreeReference parent(TreeReference baseRef) {
         if (isAbsolute()) {
-            return this;
+            return this.clone();
```
phillipm: Mimicking other similar methods that always return a copy instead of itself. Is this fine?

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -43,8 +43,9 @@
     // Multiplicity demarcates the position of a given element with respect to
     // other elements of the same name.
 
-    // Means there's a path to a node, but we haven't yet demarcated the
-    // multiplicity
+    // Since users usually want to select the first instance from the nodeset
+    // returned from a reference query, let the default multiplicity be
+    // selecting the first node.
```
phillipm: Is this comment correct?

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -629,48 +756,26 @@ public int getContext() {
...
-        //Copy construct
-        TreeReference ret = new TreeReference();
-        ret.refLevel = this.refLevel;
-        ret.contextType = this.contextType;
-        ret.instanceName = this.instanceName;
-        ret.data = new Vector<TreeReferenceLevel>();
+        TreeReference subRef = cloneWithoutLevels();
         for (int i = 0; i <= level; ++i) {
-            ret.data.addElement(this.data.elementAt(i));
+            subRef.add(this.data.elementAt(i).shallowCopy());
```
phillipm: Delegating through the 'add' method, instead of data.addElement, since it performs potentially crucial internal manipulations (hashCode = -1, size = -1, etc). 
Was there a (performance) reason that this wasn't being done before?

phillipm: We want to add this call to shallowCopy, right?

core/src/org/javarosa/core/model/instance/TreeReference.java
ctsims: Sounds correct. Alternatively if it would make this clearer we could do something like

DEFAULT_MULTIPLICITY = MULTIPLICITY_FIRST;

core/src/org/javarosa/core/model/instance/TreeReference.java
ctsims: We generally only Clone() when we modify, for performance reasons. This will work fine, I think, but not sure it's necessary.

core/src/org/javarosa/core/model/instance/TreeReference.java
ctsims: Looks right

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -629,48 +756,26 @@ public int getContext() {
...
         }
-        return ret;
-    }
-
-    public boolean hasPredicates() {
```
ctsims: Did these get removed? These are public API methods, so kind of sketch to remove them

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -221,9 +282,31 @@ public TreeReference clone() {
...
 
+    /**
+     * Return a copy of the TreeReference that doesn't include any of the
+     * TreeReferenceLevels. Useful when we are just going to overwrite the
+     * levels with new data anyways.
+     *
+     * @return a clone of this object that doesn't include any reference level
+     * data.
+     */
+    private TreeReference cloneWithoutLevels() {
```
ctsims: What does this do conceptually, exactly? only clones the instance name and root ref level? Might be a more explanatory name we can be using here.

core/src/org/javarosa/core/model/instance/TreeReference.java
phillipm: Okay, leaving the clone in.
We can refactor all the non-modification related clone at once if needed later.

core/src/org/javarosa/core/model/instance/TreeReference.java
phillipm: No, they just got moved up in the file to be close to the other predicate related methods.

core/src/org/javarosa/core/model/instance/TreeReference.java
ctsims: I think subRef.add seems like the right idea here.

Not sure about the shallow copy. The TreeReferenceLevels should basically be immutable, I'm actually not 100% sure when shallow copying them would ever be necessary....

core/test/org/javarosa/core/model/test/TreeReferenceTest.java
ctsims: /a[1]/a[5] definitely sounds like the right output there, yeah.

core/src/org/javarosa/core/model/instance/TreeReference.java
ctsims: The fix is also testing whether newRef has an existing multiplicity for this level, right?

phillipm: Pretty sure it is just to remove the whole multiplicity copying logic. The call to anchor correctly copies/merges the multiplicity values, and this call to setMultiplicity messes them up if they have the same name but are in fact different parts of the reference.

core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -337,18 +442,19 @@ else if (!this.getInstanceName().equals(contextRef.getInstanceName())) {
...
+        // apply multiplicites and fill in wildcards as necessary, based on the
+        // context ref
         for (int i = 0; i < contextRef.size() && i < newRef.size(); i++) {
-
-            //If the the contextRef can provide a definition for a wildcard, do so
-            if (TreeReference.NAME_WILDCARD.equals(newRef.getName(i)) && !TreeReference.NAME_WILDCARD.equals(contextRef.getName(i))) {
+            // If the the contextRef can provide a definition for a wildcard, do so
+            if (TreeReference.NAME_WILDCARD.equals(newRef.getName(i)) &&
+                    !TreeReference.NAME_WILDCARD.equals(contextRef.getName(i))) {
                 newRef.data.setElementAt(newRef.data.elementAt(i).setName(contextRef.getName(i)), i);
```
phillipm: I have no idea what this code is accomplishing.
As far as I can tell, wildcards in 'this.anchor(contextRef)' will either be present in contextRef too or out of scope, but shouldn't be resolved.

core/test/org/javarosa/core/model/test/TreeReferenceTest.java
```diff
@@ -136,155 +177,290 @@ public void doTest(int i) {
...
+        // ('c').contextualize('/a/*') ==> /a/*/c
+        TreeReference wildA = XPathReference.getPathExpr("/a/*").getReference();
+        contextualizeEval = floatc.contextualize(wildA);
+        assertTrue("Got " + contextualizeEval.toString() + " and expected /a/*/c" +
+                " for test of wildcard merging",
+                wildA.extendRef("c", TreeReference.INDEX_UNBOUND).equals(contextualizeEval));
+
+        // ('../*').contextualize('/a/c') ==> /a/*/
+        // XXX: what should the behavoir for this be?
+        //      /a/c or /a/*
```
phillipm: Opinions on this?

core/src/org/javarosa/core/model/instance/TreeReference.java
ctsims: Ok. It might be worth doing a profiling review of some of the treereference/general refactor work to see what the impact is. TreeReference manipulation is a pretty legit bottleneck in the code currently, so I'd wanna make sure we know what we're getting into.

core/test/org/javarosa/core/model/test/TreeReferenceTest.java
ctsims: agree that /a/*/c is the correct output

PR #73, opened 2015-04-16
Remove IDataReference

phillipm:
IDataReference was an unused abstraction, causing a bunch of unneeded indirection in the code

Crpss-request with https://github.com/dimagi/commcare-odk/pull/229


j2me/javarosa-app/build.xml
```diff
@@ -124,7 +124,6 @@
 					<parameter name="keep,allowshrinking" value="!abstract class org.javarosa.core.model.* implements org.javarosa.core.model.IFormElement" />
 					<parameter name="keep, allowshrinking" value="!abstract class org.javarosa.core.model.instance.* implements org.javarosa.core.model.IFormDataModel" />
 					<parameter name="keep ,allowshrinking" value="!abstract class org.javarosa.core.model.data.* implements org.javarosa.core.model.data.IAnswerData" />
-					<parameter name="keep,  allowshrinking" value="!abstract class org.javarosa.model.xform.* implements org.javarosa.core.model.IDataReference" />
```
phillipm: Is it okay to take this out? Have no idea what it does...

j2me/deprecated/org.javarosa.patient/src/org/javarosa/model/GraphDataGroup.java
```diff
@@ -20,7 +20,7 @@
 import java.util.Enumeration;
 import java.util.Vector;
 
-import org.javarosa.core.model.IDataReference;
+import org.javarosa.model.xform.XPathReference;
```
phillipm: Should I be concerned about modifying files in the 'deprecated' directory?

ctsims: We can likely just delete that directory now. Should add to the tech debt
roadmap, that code is dead dead.

orangejenny:
Deleting code. Glorious.

orangejenny:
looks fine, will merge as soon as it's mergeable

PR #74, opened 2015-04-20
Simple xpath expression tests while trying to reproduce bug

phillipm:
Added some more tests while trying to reproduce a bug that turned out to be a xpath syntax error.

### DIFF
--- a/core/resources/test_nested_multiplicities.xml
+++ b/core/resources/test_nested_multiplicities.xml
@@ -1,0 +1,53 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms"
+        xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+        xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>Constraints for simple and select answers</h:title>
+        <model>
+            <instance>
+                <data xmlns:jrm="http://dev.commcarehq.org/jr/xforms"
+                      xmlns="http://openrosa.org/formdesigner/21D20D14-BF66-497D-8258-8356080CDB3C"
+                      uiVersion="1" version="279" name="simple structure to test nested multiplicity queries">
+                    <show>yes</show>
+                  <bikes>
+                    <manufacturer id="surly" american="yes">
+                      <model id="karate-monkey" color="black"/>
+                      <model id="long-haul" color="black"/>
+                      <model id="cross-check" color="silver"/>
+                    </manufacturer>
+                    <manufacturer id="moots" american="yes">
+                      <model id="vamoots" color="silver"/>
+                    </manufacturer>
+                    <manufacturer id="bianchi" american="no">
+                      <model id="pista" color="seafoam"/>
+                      <model id="zolder" color="black"/>
+                    </manufacturer>
+                  </bikes>
+
+                  <silver-american>none</silver-american>
+                  <pista-color>blue</pista-color>
+                </data>
+            </instance>
+
+            <setvalue event="xforms-ready" ref="/data/silver-american"
+                value="join(' ', /data/bikes/manufacturer[@american='yes']/model[@color='silver']/@id)"/>
+            <setvalue event="xforms-ready" ref="/data/pista-color"
+                value="/data/bikes/manufacturer/model[@id='pista']/@color"/>
+        </model>
+    </h:head>
+
+    <h:body>
+        <select1 ref="/data/show">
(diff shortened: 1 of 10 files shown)


core/resources/test_nested_multiplicities.xml
```diff
@@ -0,0 +1,53 @@
...
+                      <model id="zolder" color="silver"/>
+                    </manufacturer>
+                  </bikes>
+
+                  <silver-american>none</silver-american>
+                  <pista-color>blue</pista-color>
+                </data>
+            </instance>
+
+            <setvalue event="xforms-ready" ref="/data/silver-american"
```
ctsims: are we using these anymore or are we using the inline tests? Seems confusing to include them in both places?

core/test/org/javarosa/xpath/expr/test/XPathPathExprTest.java
```diff
@@ -105,6 +115,24 @@ public void doTests() {
...
+        FormDef fd = fpi.getFormDef();
+        FormEntryModel fem = fpi.getFormEntryModel();
+
+        testEval("/data/bikes/manufacturer/model[@id='pista']/@color",
+                fd.getInstance(), null, "seafoam");
+        testEval("join(' ', /data/bikes/manufacturer[@american='yes']/model[.=1]/@id)",
+                fd.getInstance(), null, "karate-monkey vamoots");
+        // fails because of [model=1]:
+        testEval("join(' ', /data/bikes/manufacturer[@american='yes'][model=1]/model/@id)",
+                fd.getInstance(), null, new XPathTypeMismatchException());
```
ctsims: Are we 100% sure this exception type is appropriate? General XPathException might fit better?

core/src/org/javarosa/core/model/condition/EvaluationContext.java
```diff
@@ -348,11 +349,19 @@ private void expandReferenceAccumulator(TreeReference sourceRef, DataInstance so
...
                         //test the predicate on the treeElement
                         //EvaluationContext evalContext = new EvaluationContext(this, treeRef);
                         EvaluationContext evalContext = rescope(treeRef, positionContext[predIndex]);
-                        Object o = xpe.eval(sourceInstance, evalContext);
+                        Object o;
+                        try {
+                            o = xpe.eval(sourceInstance, evalContext);
+                        } catch (XPathTypeMismatchException e) {
+                            // Yay, the predicate's so jacked it failed to evaluate!
+                            passedAll = false;
```
phillipm: Should we fail silently here or not?

core/test/org/javarosa/xpath/expr/test/XPathPathExprTest.java
```diff
@@ -105,6 +115,29 @@ public void doTests() {
...
+                fd.getInstance(), null, "seafoam");
+        testEval("join(' ', /data/bikes/manufacturer[@american='yes']/model[.=1]/@id)",
+                fd.getInstance(), null, "karate-monkey vamoots");
+        testEval("join(' ', /data/bikes/manufacturer[@american='yes'][model=1]/model/@id)",
+                fd.getInstance(), null, "vamoots");
+        // if the model elements don't all have values, then the predicate
+        // doesn't evaluate correctly, and we take that as not passing,
+        // returning empty.
+        // TODO: should throw XPathExpression exception?
+        testEval("join(' ', /data/bikes/manufacturer[@american='no'][model=1]/model/@id)",
```
phillipm: Just want to check: The [model=1] predicate is nonsensical in this setting, right? Evaluating it should returns something like a nodeset, which isn't a predicate.

core/src/org/javarosa/core/model/condition/EvaluationContext.java
ctsims: Hey,

So we don't "fail" here quite yet. We only "fail" on this output if we actually try to _Access_ the nodeset. If, for instance, all we're doing is Count()ing this nodeset, we intentionally don't fail, because that can be used as a way to shield references that would otherwise crash due to unavailable data.

for example:

if(count(instance('instance')/data/node/that/may/not/exist) = 0, '', instance('instance')/data/node/that/may/not/exist)

I think we might actually need this exception to bubble all the way up for that functionality to work, thoug?

core/test/org/javarosa/xpath/expr/test/XPathPathExprTest.java
ctsims: model = 1 makes sense to me in this setting, actually. You're only including nodes into the nodeset for which model has the element value 1 (IE: <model>1</model>) and including the ID if so, so that feels like a reasonable thing to be able to do?

PR #78, opened 2015-04-24
allow custom functions to override default behavior

wpride:


wpride:
@ctsims @czue

czue:
:+1: from me but will let someone else merge

czue:
thanks for quick turnaround! what release will this be in

wpride:
hopefully 2.21 which I think is freezing today or Monday

core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -145,6 +145,13 @@ public Object eval(DataInstance model, EvaluationContext evalContext) {
 
         Hashtable funcHandlers = evalContext.getFunctionHandlers();
 
+
+        //check for custom handler, use this if it exists.
+        IFunctionHandler handler = (IFunctionHandler)funcHandlers.get(name);
+        if (handler != null) {
+            return evalCustomFunction(handler, argVals, evalContext);
```
ctsims: Question: This goes by name, what is the custom function handler is overriding a multiple argument function. 

IE: 

custom function = if(a, b, c, d) (so a ridiculous 4-if argument)

We should probably have another return path from evalCustomFunction() which says "Actually, I have a named function for this, but no matching argument # signatures" that still lets it fall back down to the default functions.

wpride:
@ctsims good call, updated. if we change the signature of the eval() method will this break anything?

core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -145,6 +145,17 @@ public Object eval(DataInstance model, EvaluationContext evalContext) {
...
+        //check for custom handler, use this if it exists.
+        try {
+            IFunctionHandler handler = (IFunctionHandler) funcHandlers.get(name);
+            if (handler != null) {
+                return evalCustomFunction(handler, argVals, evalContext);
+            }
+        } catch(XPathArityException e){
+            // we matched the name but not the arg count.
+            // continue in case the default has the right arity.
+        }
```
czue: in https://github.com/dimagi/touchforms/pull/89/files#diff-752ce0230860c47c4ce3a737af9ce26fR54 we define the return type as broadly as possible (class) so that we can return different types. will that work with this or do we need to be more specific in our overrides?

ctsims: return type should be fine as far as I know. @wpride can you confirm that'll match for this?

ctsims: @wpride ready to merge after that. Can you flag as "ready for review"?

wpride:
@ctsims @czue added unit tests, verified working after fixing a dump bug. this is gtg

core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -145,6 +145,20 @@ public Object eval(DataInstance model, EvaluationContext evalContext) {
 
         Hashtable funcHandlers = evalContext.getFunctionHandlers();
 
+        for (int i = 0; i < args.length; i++) {
+            argVals[i] = args[i].eval(model, evalContext);
+        }
```
phillipm: You shouldn't move argument evaluation in above of if/else expression short-circuiting that occurs on line 162,

core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -145,6 +145,20 @@ public Object eval(DataInstance model, EvaluationContext evalContext) {
...
+
+        //check for custom handler, use this if it exists.
+        try {
+            IFunctionHandler handler = (IFunctionHandler) funcHandlers.get(name);
+            if (handler != null) {
+                return evalCustomFunction(handler, argVals, evalContext);
+            }
+        } catch(XPathArityException e){
+            // we matched the name but not the arg count.
```
phillipm: Please add tests to verify that you actually override an existing function. Try overriding the "true" function, having it take in an argument and return "1". This is a good test, because it also checks arity errors, which you don't test for currently.

wpride:
@phillipm updated

core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -163,6 +163,20 @@ public Object eval(DataInstance model, EvaluationContext evalContext) {
...
+        try {
+            IFunctionHandler handler = (IFunctionHandler)funcHandlers.get(name);
+            if (handler != null) {
+                return evalCustomFunction(handler, argVals, evalContext);
+            }
+        } catch (XPathArityException e) {
+            // we matched the name but not the arg count. continue in case the
+            // default has the right arity, and if no default found, raise this
+            // error on exit
+            customFuncArityError = e;
```
phillipm: We need to be sure to re-raise arity exceptions if no default is found.

core/src/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -401,41 +411,59 @@ private static void checkArity(String name, int expectedArity, int providedArity
...
         Object[] typedArgs = null;
 
+        boolean argPrototypeArityMatch = false;
+        Class[] proto;
         while (typedArgs == null && e.hasMoreElements()) {
-            typedArgs = matchPrototype(args, (Class[])e.nextElement());
+            // try to coerce args into prototype, stopping on first success
+            proto = (Class[])e.nextElement();
+            typedArgs = matchPrototype(args, proto);
+            argPrototypeArityMatch = argPrototypeArityMatch ||
```
phillipm: If we don't find a prototype that is the same size as the arguments then it isn't a type mismatch, it is an arity error

PR #85, opened 2015-05-04
Have 'contextualize' copy mult info.

phillipm:
Fix contextualize to copy over multiplicity info from context reference if it will make the reference in question more specific, without overwriting existing multiplicity info.

Fixes [this ticket](http://manage.dimagi.com/default.asp?167350#932438)

### DIFF
--- a/core/src/org/javarosa/core/model/instance/TreeReference.java
+++ b/core/src/org/javarosa/core/model/instance/TreeReference.java
@@ -417,7 +417,10 @@ public class TreeReference implements Externalizable {
     }
 
     /**
-     * Evaluate this reference in terms of the base reference argument.
+     * Evaluate this reference in terms of the base reference argument. If this
+     * reference can be made more specific by filters or predicates in the
+     * context reference, it does so, but never overwrites existing filters or
+     * predicates.
      *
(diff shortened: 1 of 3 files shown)


core/src/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -460,7 +463,14 @@ public TreeReference contextualize(TreeReference contextRef) {
                 newRef.data.setElementAt(newRef.data.elementAt(i).setName(contextRef.getName(i)), i);
             }
 
-            if (!contextRef.getName(i).equals(newRef.getName(i))) {
+            if (contextRef.getName(i).equals(newRef.getName(i))) {
+                // Only copy over multiplicity info if it won't overwrite any
+                // existing preds or filters
+                if (newRef.getPredicate(i) == null &&
```
phillipm: Not sure if the 'newRef.getPredicate(i) == null' check is needed. Just leaving it in from the existing code.

ctsims: Yeah, I think that's the right play. predicates and multiplicities should be mutually exclusive, I think.

PR #91, opened 2015-05-07
Constraint args

wpride:


wpride:
@ctsims - what do you think of changing the spec to define constraints this way? arguments becoming evaluatable out of the box

wpride:
could also do help/hint text this way

core/src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1030,6 +1032,32 @@ protected QuestionDef parseControl(IFormElement parent, Element e, int controlTy
         return question;
     }
 
+    private void parseConstraintLabel(QuestionDef q, Element e) {
```
wpride: would be nice to reduce the repeated code between this and parseQuestionLabel (particularly if we use this pattern for help/hint text)

core/src/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -459,5 +459,17 @@ public void requestConstraintHint(ConstraintHint hint) throws UnpivotableExpress
...
+    /**
+     * Convenience method
+     * Get longText form of text for THIS element (if available)
+     * !!Falls back to default form if 'long' form does not exist.!!
+     * Use getSpecialFormQuestionText() if you want short form only.
+     *
+     * @return longText form
+     */
+    public String getConstraintText() {
+        return getQuestionText(getQuestion().getConstraintTextID());
```
wpride: TODO: add fallback to previous method if this doesnt exist

wpride:
@ctsims @orangejenny this is now working at least in my narrow test cases. I'm not sure about help images/audio/video at the moment - is there an example for this somewhere?

orangejenny:
This app has hint text and a bunch of media in it: https://www.commcarehq.org/a/jennytraining/apps/view/ddd0b2eb8661eba592e59d9f7e9c0528/releases/

wpride:
@orangejenny thanks. confirmed this works with help multimedia

wpride:
https://confluence.dimagi.com/display/ccinternal/Constraint+Message+Spec+Change

@ctsims per our conversation yesterday
@orangejenny fyi

ctsims:
Will - Ah, I see now. This does make sense to me and it definitely does
feel like it makes more sense for the label to be in the set of question
texts, given the MVC layout.

Couple comments:
1) we should still be namespacing the "constraint" here, since constraint
isn't in the XForms xml namespace.
2) We'll still need to continue supporting the old format of constraints as
well (we can make the parser load them in through an extra processing step
if need be). Is that supported in this PR directly? (jumping in soon).

-Clayton

wpride:
2) fixed in new commit
can you elaborate on 1?

ctsims:
Basically "constraint" isn't one of the fields in the xforms spec, so we
should theoretically be namespacing it into "our" xforms namespace

-Clayton

ctsims:
Looks good pending tests. Should test:

A) That old constraint configuration still works reliably
B) That new constraint configuration works with the same (non-arg'd) setup as (A)
C) That new constraint configuration works with arguments

wpride:
@ctsims

ctsims:
Can you replace all "help"/"label" etc bare strings with a static final thing to make sure we don't make minor mistakes based on that?

wpride:
@ctsims updated

PR #92, opened 2015-05-11
support for http://www.w3.org/2005/xpath-functions/math/

wpride:


wpride:
http://manage.dimagi.com/default.asp?138135#938211

PR #94, opened 2015-05-12
Remove title println

benrudolph:
@wpride @ctsims

### DIFF
--- a/core/src/org/javarosa/xform/parse/XFormParser.java
+++ b/core/src/org/javarosa/xform/parse/XFormParser.java
@@ -676,7 +676,6 @@ public class XFormParser {
     private void parseTitle(Element e) {
         Vector usedAtts = new Vector(); //no attributes parsed in title.
         String title = getXMLText(e, true);
-        System.out.println("Title: \"" + title + "\"");
         _f.setTitle(title);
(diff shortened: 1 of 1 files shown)


czue:
:eyes:

ctsims:
@benrudolph @wpride 

! Not 100% sure about this. this println is _super_ _super_ helpful during debugging to see what xform was being parsed when a parse exception throws,

benrudolph:
Is it possible to include the title in the exception then? I don't like the println cause we have no control over the log level and the format is useless in terms of trying to figure out the timestamp

wpride:
@ctsims my thinking was that I've always never used this sysout in debugging - I think I usually use the exception as ben suggests. If there're situations where this doesn't work or you rely on it happy to add it back in of course (though I'm definitely feeling rudolph's pain with the touchforms logs lately (which we shouldn't be completely beholden to))

ctsims:
Hey: If we can include it in traces or something that works. The thing is
that the place this is really helpful is during debugging with j2me apps
which is already incredibly painful, so I'd really _really_ like to not
lose this tool.

shouldn't touchforms be dropping stdout anyway? I'd think we'd only want to
log stderr?

PR #76, opened 2015-04-21
Handling nested predicates with relative references

phillipm:
An unfortunately large-ish PR that accomplishes the following
 - make sure (relative) predicates in triggers are contextualized correctly
 - parser warnings for control references that have predicates
 - parser warnings for binds/control elements with references pointing to non-main instances
 - cleanup of EvaluationContext.expandReferenceAccumulator
 - some tests for these changes
 - misc comments and small code refactors.

Trying to address this [ticket](http://manage.dimagi.com/default.asp?89140)

PR #97, opened 2015-05-22
Trigger related comments and field privatization

phillipm:
Method level comments relating to triggers.
Made triggerIndex private and expose it through documented methods as a means to better understand its role.

Only review after merging #76


core/src/org/javarosa/core/model/FormDef.java
```diff
@@ -98,7 +98,7 @@
     // This list is topologically ordered, meaning for any tA
     // and tB in the list, where tA comes before tB, evaluating tA cannot
     // depend on any result from evaluating tB
-    public Vector<Triggerable> triggerables;
+    private Vector<Triggerable> triggerables;
```
phillipm: Don't pull until I add a commit to schema-gen code that makes use of this being public.

phillipm: Good to go now.

phillipm:
Can't figure out why this PR is messed up
Cherry-picked relevant commits into this PR https://github.com/dimagi/javarosa/pull/105

PR #99, opened 2015-05-28
Support pixel-based case list column widths

orangejenny:
http://manage.dimagi.com/default.asp?169125

J2ME doesn't actually support pixel-based columns, but it should still handle percentage-based ones.

Cross-requested with commcare and commcare-odk


orangejenny:
https://github.com/dimagi/commcare-odk/pull/314
https://github.com/dimagi/commcare-odk/pull/314

PR #101, opened 2015-06-09
add old J2ME functions

wpride:


wpride:
J2ME needs this, hence the Jenkins failure. I figured I'd just re-add these instead of diving into the j2me codebase since that's mostly stagnant anyways.

PR #103, opened 2015-06-15
use the old code to fallback more elegantly

wpride:
2.22 bug fix where we were crashing when the constraint message was being retrieved when it wasn't set. the existing code already had the logic for this, I just wasn't polymorphing correctly

http://manage.dimagi.com/default.asp?171245

### DIFF
--- a/core/src/org/javarosa/form/api/FormEntryPrompt.java
+++ b/core/src/org/javarosa/form/api/FormEntryPrompt.java
@@ -193,11 +193,22 @@ public class FormEntryPrompt extends FormEntryCaption {
         }
     }
 
+    public String getConstraintText(){
+        return getConstraintText(null);
+    }
+
     public String getConstraintText(IAnswerData attemptedValue) {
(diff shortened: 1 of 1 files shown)


core/src/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -193,11 +193,22 @@ public String getAnswerText() {
...
+    }
+
     public String getConstraintText(IAnswerData attemptedValue) {
         return getConstraintText(null, attemptedValue);
     }
 
     public String getConstraintText(String textForm, IAnswerData attemptedValue) {
+        // new constraint spec uses "alert" form XForm spec 8.2.4
+        // http://www.w3.org/TR/xforms/#ui-commonelems
+        String newConstraintMsg =  this.localizeText(getQuestion().getQuestionString(XFormParser.CONSTRAINT_ELEMENT));
```
ctsims: This doesn't look like it handles the situation where the textForm is relevant.

PR #105, opened 2015-06-19
Comments and field privatization for triggers

phillipm:
An old PR that I had to recreate because of merging issues:

Method level comments relating to triggers.
Made triggerIndex private and expose it through documented methods as a means to better understand its role.

### DIFF
--- a/core/src/org/javarosa/core/model/FormDef.java
+++ b/core/src/org/javarosa/core/model/FormDef.java
@@ -545,8 +545,6 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
 
     /**
      * Add a Condition to the form's Collection.
-     *
-     * @param condition the condition to be set
      */
     public Triggerable addTriggerable(Triggerable t) {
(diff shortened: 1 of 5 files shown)


core/src/org/javarosa/core/model/FormDef.java
```diff
@@ -109,10 +109,19 @@
     // arguments to captions
     private Vector outputFragments;
 
-    public Hashtable<TreeReference, Vector<Triggerable>> triggerIndex;
+    /**
+     * Map references to the calculate/relevancy conditions that depend on that
+     * reference's value. Used to trigger re-evaluation of those conditionals
+     * when the reference is updated.
+     */
+    private Hashtable<TreeReference, Vector<Triggerable>> triggerIndex;
```
phillipm: Made this private and exposed methods to access its data in an immutable manner.

phillipm:
@ctsims Can you look at this? It should be a quick review but is worth checking that my trigger-related comments are correct.

PR #108, opened 2015-06-22
Migrating to gradle build

phillipm:
Have main JavaRosa sources be built by gradle. Left the j2me building with Ant.

This doesn't accomplish too much in and of itself, but the next PR migrates test infrastructure to Junit4, which when coupled with gradle makes for easier testing (i.e. to run a single test file: gradle test --tests "*XPathPathExprTest" --info).

Once this is merged we will need to adapt the [javarosa-core-library](http://jenkins.dimagi.com/job/javarosa-core-library) build to mimic the [javarosa-core-library-gradle](http://jenkins.dimagi.com/job/javarosa-core-library-gradle) build. Devs will need to update their Android Studio projects by removing the current javarosa module and importing javarosa/core (Clayton, not sure how this works with Eclipse...).

Directory changes to implement gradle's convention: 
 - move core/src -> core/src/main/java
 - move core/test -> core/src/test/java


snopoke:
Oooo :clap:

PR #109, opened 2015-06-22
Migrate tests to use JUnit4

phillipm:
I got tired of J2MEs dated version of JUnit.

This enables easier command-line and IDE test invocations.

This creates some problems with TextFormTests and XPathParseTest that I will fix as soon as this gets merged.

This will require adding JUnit4 as a lib dependency in Android Studio.

**Shares commits with https://github.com/dimagi/javarosa/pull/108.**


core/src/test/java/org/javarosa/xpath/test/XPathParseTest.java
```diff
@@ -209,34 +191,26 @@
...
-
-    public XPathParseTest() {
-        super();
-    }
-
-    public Test suite() {
-        TestSuite aSuite = new TestSuite();
+    // TODO PLM: fix this!
+    /*
+    public static TestSuite suite() {
```
phillipm: Note that I need to fix this soon.

core/src/test/java/org/javarosa/core/form/api/test/TextFormTests.java
```diff
@@ -17,13 +17,11 @@
...
-import j2meunit.framework.TestCase;
-import j2meunit.framework.TestMethod;
-import j2meunit.framework.TestSuite;
-
-public class TextFormTests extends TestCase {
+import org.junit.Test;
+import org.junit.Before;
+import static org.junit.Assert.*;
 
+public class TextFormTests {
```
phillipm: The tests in this file were never being called! Enabling them, it looks like 3 of them are failing

core/src/test/java/org/javarosa/core/form/api/test/TextFormTests.java
```diff
@@ -288,6 +229,4 @@ public void testPromptsNoLocalizer() {
             fail("Help text getter/setter broken");
```
phillipm: Fails with a null pointer exception

core/src/test/java/org/javarosa/core/form/api/test/TextFormTests.java
```diff
@@ -279,8 +220,8 @@ public void testPromptIDsNoLocalizer() {
         }
```
phillipm: Fails due to equality check in between String and QuestionString objects

core/src/test/java/org/javarosa/core/form/api/test/TextFormTests.java
```diff
@@ -155,9 +92,9 @@ public void testTextForms() {
             fail("getLongText() not falling back to default text form correctly");
```
phillipm: Fails, due to test resource input not matching expected behavior

ctsims: Hey,

These are fixed in the 2.22 branch.

core/src/test/java/org/javarosa/core/form/api/test/TextFormTests.java
phillipm: Clayton fixed this, nv

PR #114, opened 2015-07-10
Fix NullPointerException in genStyleFromHints

orangejenny:
http://manage.dimagi.com/default.asp?176010


phillipm:
Does this need to also go into oldui/commcare_2.22?

orangejenny:
@phillipm There's an oldui branch for javarosa?

phillipm:
@orangejenny Good point, there isn't!

PR #116, opened 2015-07-13
Remove generics annotations because of J2ME

phillipm:
J2ME compatibility fix for https://github.com/dimagi/javarosa/pull/105

### DIFF
--- a/core/src/org/javarosa/xform/parse/XFormParser.java
+++ b/core/src/org/javarosa/xform/parse/XFormParser.java
@@ -2796,8 +2796,8 @@ public class XFormParser {
         Vector edges = new Vector();
 
         //build graph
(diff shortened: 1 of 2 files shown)


core/src/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2796,8 +2796,8 @@ private void checkDependencyCycles() {
         Vector edges = new Vector();
 
         //build graph
-        for (Enumeration e = _f.refWithTriggerDependencies(); e.hasMoreElements(); ) {
-            TreeReference trigger = (TreeReference)e.nextElement();
+        for (Enumeration<TreeReference> e = _f.refWithTriggerDependencies(); e.hasMoreElements(); ) {
```
phillipm: Are annotations here accepted?

PR #118, opened 2015-07-14
Move form instance loader logic from ODK

phillipm:
Moving code that loads form instances from ODK into JavaRosa so that it can be reused by debugging utils.

Cross-requested with https://github.com/dimagi/commcare-odk/pull/418

### DIFF
--- a/core/src/org/javarosa/core/util/Pair.java
+++ b/core/src/org/javarosa/core/util/Pair.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (C) 2014 Open WhisperSystems
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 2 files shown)


core/src/org/javarosa/core/util/Pair.java
```diff
@@ -0,0 +1,57 @@
...
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
phillipm: Just here for review, don't pull into repo due to licensing

PR #120, opened 2015-07-14
Extend CLI interface to load form instances

phillipm:
Enable JavaRosa form runner CLI to load particular form instances
Try it out with: 
```
java -jar form_translate.jar runinstance form.xml instance.xml
```

### DIFF
--- a/util/schema-gen/src/org/javarosa/xform/schema/FormInstanceLoader.java
+++ b/util/schema-gen/src/org/javarosa/xform/schema/FormInstanceLoader.java
@@ -6,36 +6,58 @@ import org.javarosa.core.model.instance.InstanceInitializationFactory;
 import org.javarosa.core.model.instance.TreeElement;
 import org.javarosa.core.model.instance.TreeReference;
 import org.javarosa.form.api.FormEntryModel;
+import org.javarosa.xform.parse.XFormParseException;
 import org.javarosa.xform.parse.XFormParser;
 import org.javarosa.xform.util.XFormUtils;
 
+import java.io.IOException;
 import java.io.InputStream;
 
 /**
+ * Logic for loading a particular xml form instance into a FormDef.
+ *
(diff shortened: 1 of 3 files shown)


util/schema-gen/src/org/javarosa/xform/schema/FormInstanceLoader.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Logic for loading a particular xml form instance into a FormDef.
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class FormInstanceLoader {
+    private static InstanceInitializationFactory iif;
```
ctsims: How does this get set?

We'll need to be able to define this for an XForm that just got parsed to be properly initialized.

ctsims:
@phillipm looks good other than this one area.

util/schema-gen/src/org/javarosa/xform/schema/FormInstanceLoader.java
phillipm: No idea, it was unset in FormLoaderTask: https://github.com/dimagi/commcare-odk/blob/master/app/src/org/odk/collect/android/tasks/FormLoaderTask.java#L170

It looks like in this particular case the XFormPlayer already does the iif setup by calling XFormEnvironment.setup() and using https://github.com/dimagi/javarosa/blob/master/util/schema-gen/src/org/javarosa/engine/MockupProviderFactory.java

ctsims: This variable is currently unused in this setup, then. It's private here and unset.

phillipm:
@ctsims I removed the InstanceInitializationFactory since FormDef.initialize is called later anyways by the XFormPlayer.
Ready to look at again.

PR #122, opened 2015-07-15
Remove no-op equiv piece of logic in TreeElement.populate

phillipm:
Remove conditionals from TreeElement.populate that have test expressions that are never satisfied.

### DIFF
--- a/core/src/org/javarosa/core/model/instance/TreeElement.java
+++ b/core/src/org/javarosa/core/model/instance/TreeElement.java
@@ -70,7 +70,7 @@ public class TreeElement implements Externalizable, AbstractTreeElement<TreeElem
     //take up a huuuge amount of space together.
     private Vector observers = null;
     protected Vector<TreeElement> attributes = null;
-    protected Vector children = null;
+    protected Vector<TreeElement> children = null;
 
     /* model properties */
     protected int dataType = Constants.DATATYPE_NULL; //TODO
@@ -123,8 +123,6 @@ public class TreeElement implements Externalizable, AbstractTreeElement<TreeElem
      * Construct a TreeElement which represents an attribute with the provided
      * namespace and name.
      *
-     * @param namespace
-     * @param name
      * @return A new instance of a TreeElement
      */
     public static TreeElement constructAttributeElement(String namespace, String name) {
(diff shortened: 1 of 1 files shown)


core/src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -936,56 +931,16 @@ public void populate(TreeElement incoming, FormDef f) {
...
-            // make sure ordering is preserved (needed for compliance with xsd schema)
-            if (this.getNumChildren() != names.size()) {
-                throw new RuntimeException("sanity check failed");
-            }
-
-            for (int i = 0; i < this.getNumChildren(); i++) {
-                TreeElement child = this.getChildAt(i);
-                String expectedName = (String)names.elementAt(i);
-
-                if (!child.getName().equals(expectedName)) {
```
phillipm: The test expression in this conditional is never true!
Hence, remove the conditional altogether.

core/src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -936,56 +931,16 @@ public void populate(TreeElement incoming, FormDef f) {
...
 
-            //We used to wipe out non-repeated elements, but we won't anymore
-            //since we commonly use the pattern.
-            for (int i = 0; i < this.getNumChildren(); i++) {
-                TreeElement child = this.getChildAt(i);
-                names.addElement(child.getName());
-            }
-
-            // make sure ordering is preserved (needed for compliance with xsd schema)
-            if (this.getNumChildren() != names.size()) {
```
phillipm: This is always false...

core/src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -936,56 +931,16 @@ public void populate(TreeElement incoming, FormDef f) {
...
-                    }
-
-                    this.removeChildAt(j);
-                    if (children == null) {
-                        children = new Vector();
-                    }
-                    this.children.insertElementAt(child2, i);
-                }
-            }
-            // java i hate you so much
```
phillipm: :)

PR #123, opened 2015-07-15
Cleanup of comments & unused args while working on Form Instance Loading

phillipm:
- Removed javadoc method params that didn't have comments associated with them
 - Removed unused arguments from methods

Cross-request: https://github.com/dimagi/commcare-odk/pull/419

__Only review after merging https://github.com/dimagi/javarosa/pull/122__

### DIFF
--- a/core/src/org/javarosa/core/model/Action.java
+++ b/core/src/org/javarosa/core/model/Action.java
@@ -1,6 +1,3 @@
-/**
- *
- */
 package org.javarosa.core.model;
 
 import java.io.DataInputStream;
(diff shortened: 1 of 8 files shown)


ctsims:
@phillipm Hey, this looks fine to me. ready to pull.

One question: How broad of a net are you casting when you change method signatures? I don't believe it to be the case here, but a couple recent refactors broke, for instance, the CommCare and JavaRosa utility libraries used for the CLI, just wanted to be sure that we knew what needed to be included when library code signatures are changed.

phillipm:
Thanks for checking. When I remember I grep of all 3 repos and all their util builds (outside of the IDE). In this case I had forgotten to do that; luckily it looks like I caught all instances of 'populate'.

Should be good to go (other than needing to review/merge https://github.com/dimagi/javarosa/pull/122 first)

PR #124, opened 2015-07-15
Form instance loading doesn't handle leaf attributes

phillipm:
Form Instance loading doesn't correctly copy element attributes for leaf nodes.

Fix for http://manage.dimagi.com/default.asp?170392

__Only review/merge after https://github.com/dimagi/javarosa/pull/123 is merged__

### DIFF
--- a/core/src/org/javarosa/core/model/instance/TreeElement.java
+++ b/core/src/org/javarosa/core/model/instance/TreeElement.java
@@ -922,8 +922,7 @@ public class TreeElement implements Externalizable, AbstractTreeElement<TreeElem
      */
     public void populate(TreeElement incoming) {
         if (this.isLeaf()) {
-            // check that incoming doesn't have children?
-
+            // copy incoming element's value over
             IAnswerData value = incoming.getValue();
             if (value == null) {
                 this.setValue(null);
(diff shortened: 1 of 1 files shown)


core/src/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -998,25 +953,27 @@ public void populate(TreeElement incoming, FormDef f) {
...
-            }
+        }
+
+        // copy incoming element's attributes over
+        for (int i = 0; i < incoming.getAttributeCount(); i++) {
+            String name = incoming.getAttributeName(i);
+            String ns = incoming.getAttributeNamespace(i);
+            String value = incoming.getAttributeValue(i);
+
+            this.setAttribute(ns, name, value);
```
phillipm: The fix: moving this loop out of the else branch of the conditional above, which checks whether a node is a leaf. Attribute copying should happen regardless, right?

phillipm:
@ctsims This is pretty scary! This bug has been around for 3 years; is it really possible that it only came up now?

phillipm:
Also, this fix should probably go into 2.22, but it is hard for me to know the repercussions of this code change, especially since there are no tests around this part of the code. 
I plan to write tests soon, now that I added infrastructure to JavaRosa to load form instances.

czue:
woot!

PR #131, opened 2015-07-28
Gradle harness

wpride:
Update gradle build to also build schema-gen harness jar (into build/libs/harnes.jar)

Cross-requested with https://github.com/dimagi/commcare-odk/pull/471


wpride:
allows us to build the CLI code within gradle

PR #133, opened 2015-08-03
Add code coverage plugin to gradle

phillipm:
To invoke it: ```gradle check jacocoTestReport```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ pmd {
 }
 
 jacoco {
-    toolVersion = "0.7.4"
+    toolVersion = "0.7.4.+"
 }
 
 test {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -94,7 +99,7 @@ the path of the library archive (make sure you are using the version of the
 archive compatible with the version you wish to build) */
 task extractLibs(type: Copy) {
     if (buildDeps != null) {
-        from zipTree(file("${buildDeps}")) into file("${projectDir.parent}")
+        from zipTree(file("${buildDeps}")) into file("${projectDir}")
```
phillipm: Needed when we moved the build file out of core.

build.gradle
```diff
@@ -25,6 +26,10 @@ pmd {
     sourceSets = [sourceSets.main]
 }
 
+jacoco {
+    toolVersion = "0.7.4.+"
```
phillipm: 0.7.5 doesn't work with jenkins yet.

PR #137, opened 2015-08-07
Updated javarosa to include binary dependencies

ctsims:


snopoke:
Woop, was just thinking we should do this yesterday.

PR #146, opened 2015-08-19
Don't Pull

wpride:


wpride:
@ctsims  this caching is causing issues again, this time on touchforms. When you try to delete a repeat that's in the middle of a repeat group, this is returning the cached value of the deleted presumably because the hash is the same for the deleted item and the new one that's taking its place. Does J2ME already solve this bug in some clever way? Surprised it hasn't manifested over there at all. One possible solution would be to manually remove the caching when the repeat is deleted

@benrudolph

benrudolph:
instinctively it seems like the easiest solution would be to bust the cache on deletion. wouldn't have that many side effects and doesn't seem like it'd be too complicated. don't have good context on it though

wpride:
@benrudolph agreed that seems easiest, however at the moment there's no functionality to manually deregister a key or even but the cache (so not something we can just do from jython short of creating a new DataInstance). It also looks like there's a cache cleanup task that (maybe?) should be catching the deletion so wonder if there's a bug there... unfortunately the code is not super legible

ctsims:
Hey,

Definitely agree that busting the cache seems like the right idea. would be good to debug why that's not getting triggered.

PR #147, opened 2015-08-20
fix delete repeat bug

wpride:
@ctsims @benrudolph this fixes the issue by just wiping the cache when we delete a repeat. this happens rarely enough (not at all in ODK) that I think this is acceptable for now. Didn't want to spend too much time on this since it'll (hopefully) be addressed when we refactor the session into java

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/instance/DataInstance.java
+++ b/core/src/main/java/org/javarosa/core/model/instance/DataInstance.java
@@ -320,4 +320,8 @@ public abstract class DataInstance<T extends AbstractTreeElement<T>> implements 
     public void setCacheHost(CacheHost cacheHost) {
         this.mCacheHost = cacheHost;
     }
+
+    public void cleanCache() {
+        referenceCache = new CacheTable<TreeReference, T>();
+    }
 }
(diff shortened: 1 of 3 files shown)


benrudolph:
:tada: will let @ctsims do the mergin

core/src/main/java/org/javarosa/core/model/instance/DataInstance.java
```diff
@@ -320,4 +320,8 @@ public CacheHost getCacheHost() {
     public void setCacheHost(CacheHost cacheHost) {
         this.mCacheHost = cacheHost;
     }
+
+    public void cleanCache() {
+        referenceCache = new CacheTable<TreeReference, T>();
```
ctsims: Can you nuke this cache by doing, like, "referenceCache.clear" or something? These are somewhat sketchy as they run a cleanup thread against their objects, so it's a bit dangerous to reassign.

wpride:
@benrudolph

benrudolph:
:dancers: :dancers: yes! this worked!

wpride:
:fire:

PR #148, opened 2015-08-21
Image capture resizing

amstone326:
Cross: https://github.com/dimagi/commcare-odk/pull/549


core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -324,20 +327,24 @@ private void initState() {
 
     public XFormParser(Reader reader) {
         _reader = reader;
+        extensionParsers = new Vector<>();
```
phillipm: Not sure if J2ME supports empty <>'s. Might want to check or explicitly write out ```<QuestionExtensionParser>```

phillipm:
Looks good to me, other than double checking J2ME generics compatibility

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
amstone326: Good call, pretty sure you're right.

core/src/main/java/org/javarosa/core/model/QuestionDef.java
```diff
@@ -240,12 +243,10 @@ public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOExce
...
         setDynamicChoices((ItemsetBinding)ExtUtil.read(dis, new ExtWrapNullable(ItemsetBinding.class)));
         mQuestionStrings = (Hashtable<String, QuestionString>)ExtUtil.read(dis, new ExtWrapMap(String.class, QuestionString.class));
+        extensions = (Vector)ExtUtil.read(dis, new ExtWrapListPoly(), pf);
     }
 
-    /*
-     * (non-Javadoc)
-     * @see org.javarosa.core.util.Externalizable#writeExternal(java.io.DataOutputStream)
-     */
+    @Override
```
ctsims: You can't do @override declarations in the javarosa codbase due to 1.3 compliance

core/src/main/java/org/javarosa/xform/parse/QuestionExtensionParser.java
```diff
@@ -0,0 +1,36 @@
...
+ * parsing
+ *
+ * @author amstone
+ */
+public abstract class QuestionExtensionParser {
+
+    // The name of the question type that this parser is for ("input", "upload", etc.)
+    private String elementName;
+
+    public QuestionExtensionParser() {
```
ctsims: Empty constructor shouldn't be necessary here.

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -910,6 +917,7 @@ protected QuestionDef parseUpload(IFormElement parent, Element e, int controlUpl
         Vector usedAtts = new Vector();
         usedAtts.addElement("mediatype");
         usedAtts.addElement(REF_ATTR);
+        usedAtts.addElement("imageDimensionScaledMax");
```
ctsims: This should probably be in the extension itself, shouldn't it?

core/src/main/java/org/javarosa/xform/util/XFormUtils.java
```diff
@@ -61,6 +63,39 @@ public static FormDef getFormRaw(InputStreamReader isr) throws XFormParseExcepti
         return _factory.getXFormParser(isr).parse();
     }
 
+    public static FormDef getFormFromInputStream(InputStream is, QuestionExtensionParser extensionParser)
```
ctsims: I'd guess we want this to be a list or a vector to make it easier to add a bunch of these.

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
amstone326: The only way that would really work would be to have parseControl return the used attributes for whatever extension parser was parsed within it, since there's no interaction with the extensions in this method. Are we ok with the awkwardness of parseControl() having to return like Pair<QuestionDef, List<UsedAtts>>?

phillipm: Heads up that Pair isn't J2ME compatible :(

core/src/main/java/org/javarosa/core/model/ImageRestrictionExtension.java
```diff
@@ -0,0 +1,43 @@
...
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+/**
+ * Represents an extension to the "upload" question type, for setting a maximum
+ * dimension for the size of an image capture
+ *
+ * @author amstone
+ */
+public class ImageRestrictionExtension implements QuestionDataExtension {
```
ctsims: One question here: Do we want to make this slightly more expressive? IE should this be something like:

ImageCaptureExtension

or something that makes it clear that it's cool for people to continue adding other image question data here?

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -324,20 +327,24 @@ private void initState() {
 
     public XFormParser(Reader reader) {
         _reader = reader;
+        extensionParsers = new Vector<QuestionExtensionParser>();
```
ctsims: These can be moved to the private Vector<> definition above for clarity/lack of repeating

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2015,7 +2035,7 @@ private FormInstance parseInstance(Element e, boolean isMainInstance) {
             instanceModel.setName(name);
         }
 
-        Vector usedAtts = new Vector();
+        Vector<String> usedAtts = new Vector<String>();
```
ctsims: Thanks for these!

core/src/main/java/org/javarosa/xform/util/XFormUtils.java
```diff
@@ -61,6 +63,42 @@ public static FormDef getFormRaw(InputStreamReader isr) throws XFormParseExcepti
         return _factory.getXFormParser(isr).parse();
     }
 
+    public static FormDef getFormFromInputStream(InputStream is,
+                                                          Vector<QuestionExtensionParser> extensionParsers)
```
ctsims: super minor point: If these are read-only depending on the circumstance it might be easier to create them as an Array rather than a vector so the declaration is

getFormFromInputStream(is, new QuestionExtensionParser[]{myparser, myotherparser})

rather than

Vector extensions = new Vector()
extensions.add(myparser)
extensions.add(otherparser)
getFormFromInputStream(is, extensions)

but feel free to ignore (IE: No strong preference)

core/src/main/java/org/javarosa/core/model/ImageRestrictionExtension.java
amstone326: Ah yea good point

ctsims:
Good to go. would like a test that parses a form with the upload extension and gets the max size, though

amstone326:
Test added.

core/src/test/java/org/javarosa/core/model/test/UploadExtensionTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Before
+    public void init() {
+        extensionParsers = new Vector<>();
+        extensionParsers.add(new UploadQuestionExtensionParser());
+    }
+
+    @Test
+    public void testParseMaxDimenWithPx() {
+        FormParseInit formWithPx = new FormParseInit("/test_upload_extension_1.xml", extensionParsers);
```
ctsims: Now that we're expanding a lot, can we add these to subfolders like

/xform_tests/test_upload...
?

core/src/test/resources/test_upload_extension_2.xml
```diff
@@ -0,0 +1,25 @@
...
+            <bind nodeset="/data/image_capture" type="binary"/>
+            <itext>
+                <translation lang="en" default="">
+                    <text id="image_capture-label">
+                        <value>Image Capture</value>
+                    </text>
+                </translation>
+            </itext>
+            <setvalue event="xforms-ready" ref="/data/meta/deviceID" value="instance('commcaresession')/session/context/deviceid"/><setvalue event="xforms-ready" ref="/data/meta/timeStart" value="now()"/><bind nodeset="/data/meta/timeStart" type="xsd:dateTime"/><setvalue event="xforms-revalidate" ref="/data/meta/timeEnd" value="now()"/><bind nodeset="/data/meta/timeEnd" type="xsd:dateTime"/><setvalue event="xforms-ready" ref="/data/meta/username" value="instance('commcaresession')/session/context/username"/><setvalue event="xforms-ready" ref="/data/meta/userID" value="instance('commcaresession')/session/context/userid"/><setvalue event="xforms-ready" ref="/data/meta/instanceID" value="uuid()"/><setvalue event="xforms-ready" ref="/data/meta/appVersion" value="instance('commcaresession')/session/context/appversion"/></model>
```
ctsims: Can we strip meta stuff? makes it a bit clearer.

amstone326:
Ready again

PR #149, opened 2015-08-25
Remove StorageFullException!

phillipm:
@ctsims says ```StorageFullException``` is completely useless, so let's take it out

Cross-requested with:
https://github.com/dimagi/commcare/pull/149
https://github.com/dimagi/commcare-odk/pull/550


j2me/core/src/org/javarosa/j2me/storage/rms/RMSStorageUtility.java
```diff
@@ -248,7 +246,7 @@ public void write (Persistable p) throws StorageFullException {
             if (!setReserveBuffer(bytesNeededEstimate)) {
                 setClean();
                 //can't reserve needed space to update index
-                throw new StorageFullException();
+                throw new RuntimeException();
```
phillipm: Seem alright to throw ```RuntimeException``` in place of ```StorageFullException```?

amstone326: Seems ok, but would probably be good to include a message

ctsims: I think we may catch this upstream to do something specific, would be good to double check?

amstone326: Is that true for everywhere that we're replacing StorageFullException with RuntimeException? Just seems like if we're going to replace StorageFullExceptions with something more generic, it can't hurt to give some slightly descriptive message.

phillipm:
Ready to go again. What I realized is that `StorageFullException` extends `RuntimeException`, hence it doesn't ever need to be in `throws` method qualifiers, nor do you need statements that catch it and throw a new `RuntimeException`. So I'm remove such uses of `StorageFullException` but leaving the rest in.

PR #150, opened 2015-08-25
Add first & last functions

orangejenny:


ctsims:
oh, Question: Did you implement this as 

last(/nodeset/path)
?

That's actually not the semantics for the last method.

last should be implemented as a context aware counter that is based on the current predicate set, so insetad of

last(/nodeset/value)

you want

/nodeset/value[last()]

orangejenny:
Closing this - the use case I wanted to address doesn't match XPath's actual first/last semantics.

PR #154, opened 2015-08-26
Remove IXFormParserFactory interface

phillipm:
IXFormParserFactory was only implemented once and never used in the way that interfaces are useful, so let's remove it.

Also remove unused versions of ```getXFormParser```


ctsims:
Phillip:

Did you check the getXFormParser stuff against J2ME's client?

phillipm:
@ctsims yup, but thanks for double checking!

PR #155, opened 2015-08-26
Use a hardcoded template for casedb instances

phillipm:
Change ```DataInstance.initialize``` to return a DataInstance object. When dealing with an ```ExternalDataInstance``` the ```InstanceInitiliazationFactory``` being used should perform specialization via the ```getSpecializedExternalDataInstance``` implementation. For instance, if dealing with an instance with the 'casedb' id, getSpecializedExternalDataInstance should create a CaseDataInstance instance from the provided ExternalDataInstance.

Contains commits from https://github.com/dimagi/javarosa/pull/154

Cross-requested with 
https://github.com/dimagi/commcare/pull/155
https://github.com/dimagi/commcare-odk/pull/560

Regarding serialization stuff in [CoreModelModule](https://github.com/dimagi/javarosa/blob/master/core/src/main/java/org/javarosa/core/model/CoreModelModule.java#L35), do I need to be concerned with the fact that I've created a class that extends ```ExternalDataInstance``` (in the commcare repo)?


core/src/test/java/org/javarosa/core/model/instance/test/DataInstanceTest.java
```diff
@@ -24,7 +24,7 @@
 
 public class DataInstanceTest {
 
-    private static final String formPath = new String("/test_xpathpathexpr.xml");
```
phillipm: Back when I was first started writing java code :)

amstone326: Amazing

amstone326: The actual scary part here is that this made it through code review?..

core/src/main/java/org/javarosa/core/model/instance/ExternalDataInstanceFactory.java
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Dispatchs on instance ID to create new external data instances.  Useful for
+ * injecting custom functionality for specific data instances, for instance if
+ * you want casedb template resolution to behave different than normal external
+ * instance resolution.
+ *
+ * @author Phillip Mates (pmates@dimagi.com)
+ */
+public class ExternalDataInstanceFactory {
+    private static final Hashtable<String, ExternalDataInstanceBuilder> instanceIdToBuilder =
```
ctsims: Hm, I would _really_ prefer to not increase the number of static builders/factories around the code. They are incredibly hard to sandbox and really harm the modularity of the code.

Can we make this a factory that we register with the static session or IIF or something? We already have an appealing entry point for when this would be injected.

phillipm:
@ctsims Thanks for the pointers on avoiding static registration/dispatch. This is good to go again.

core/src/main/java/org/javarosa/xform/parse/XFormParserFactory.java
```diff
@@ -27,32 +27,11 @@ public XFormParserFactory(Interner<String> stringCache) {
 
     public XFormParser getXFormParser(Reader reader) {
         XFormParser parser = new XFormParser(reader);
-        init(parser);
-        return parser;
-    }
 
-    private void init(XFormParser parser) {
```
ctsims: are we sure none of this affected J2ME stuff, out of curiosity? Is there a way to run a quick Jenkins test or something?

phillipm: I did a grep, there are only two uses and they are in javarosa/core/src/main/java/org/javarosa/xform/util/XFormUtils.java

ctsims: cool!

PR #156, opened 2015-09-01
fix constraint pivot

wpride:
http://manage.dimagi.com/default.asp?176697#982290

If there is more than one pivot (IE both a > and <) we should just not use the hint because they're likely opposing

### DIFF
--- a/core/src/test/resources/test_constraints.xml
+++ b/core/src/test/resources/test_constraints.xml
@@ -1,0 +1,56 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml"
+        xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+        xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>Constraints for simple and select answers</h:title>
+        <model>
+            <instance>
+                <data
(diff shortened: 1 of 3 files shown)


ctsims:
Will:

Can you add a bunch of tests around the pivots that illustrate the thing that was breaking?

wpride:
@ctsims

core/src/test/resources/test_constraints.xml
```diff
@@ -0,0 +1,56 @@
...
+                    <constraint-min />
+                    <constraint-max-and-min />
+                    <constraint-max-or-min />
+                </data>
+            </instance>
+
+            <bind nodeset="/data/constraint-max" constraint=". < 30" type="xsd:int"/>
+            <bind nodeset="/data/constraint-min" constraint=". > 10" type="xsd:int"/>
+            <bind nodeset="/data/constraint-max-and-min" constraint=". < 100 and . > 300" type="xsd:int"/>
+            <bind nodeset="/data/constraint-max-or-min" constraint=". < 30 or . < 100" type="xsd:int"/>
```
ctsims: This is actually "min or min"

ctsims:
And to confirm, this does (unfortunately) mean that we lose ".> 0 .< 100"
right? That definitely used to work, and must have broken at some point so
it's better to disable it if so, just unfortunate to lose it, since that's
actually about 90% of what hte pivots do (if there's no minimum bound we
can't constrain the input size)

wpride:
fixed and confirmed

PR #161, opened 2015-09-04
Fix for multiselect questions in CLI

ctsims:
CLI Didn't used to support multiselect questions correctly

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/data/SelectMultiData.java
+++ b/core/src/main/java/org/javarosa/core/model/data/SelectMultiData.java
@@ -20,7 +20,7 @@ import java.util.Vector;
  * @author Drew Roos
  */
 public class SelectMultiData implements IAnswerData {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/javarosa/core/model/data/SelectMultiData.java
```diff
@@ -58,7 +58,7 @@ public void setValue(Object o) {
      * (non-Javadoc)
      * @see org.javarosa.core.model.data.IAnswerData#getValue()
      */
-    public Object getValue() {
+    public Vector<Selection> getValue() {
```
phillipm: Can't remember, will J2ME crash with this generics type annotation?

ctsims: Nope, since we started using the 1.5 weaver, generics are a go in j2me code.

PR #168, opened 2015-10-01
Override hashCode whenever equals is overridden

phillipm:
Due to the interdependence of the `hashCode` and `equals` contracts (see [hashCode docs](https://docs.oracle.com/javase/1.5.0/docs/api/java/lang/Object.html#hashCode%28%29)), whenever one is overridden, it is pretty important to override the other.

For all the implementers of `XPathExpression` this was being done, but in a very round about way via the `toString` implementation. It is important that the fields used to calculate `equals` are also used in determining the result of `hashCode` so the distance between the two implementations could lead to bugs.

This PR enforces that all implementers of `XPathExpression` explicitly override both `hashCode` and `equals`.

In addition, I removed some duplicate code between the implementations of `XPathBinaryOpExpr` by consolidating the `op` field from all of the implementers into `XPathBinaryOpExpr`. Note that I took care to not change the behaviour of `read/writeExternal`, even though they have been refactored.

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
+++ b/core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
@@ -581,17 +581,13 @@ public class TreeReference implements Externalizable {
         return false;
     }
 
+    @Override
     public int hashCode() {
         if (hashCode != -1) {
             return hashCode;
         }
-        int hash = (DataUtil.integer(refLevel)).hashCode();
+        int hash = refLevel;
         for (int i = 0; i < size(); i++) {
-            //NOTE(ctsims): It looks like this is only using Integer to
-            //get the hashcode method, but that method
-            //is just returning the int value, I think, so
-            //this should potentially just be replaced by
-            //an int.
             Integer mult = DataUtil.integer(getMultiplicity(i));
             if (i == 0 && mult.intValue() == INDEX_UNBOUND)
                 mult = DataUtil.integer(0);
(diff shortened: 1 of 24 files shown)


core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -581,32 +581,29 @@ public boolean equals(Object o) {
...
+        int hash = refLevel;
         for (int i = 0; i < size(); i++) {
-            //NOTE(ctsims): It looks like this is only using Integer to
-            //get the hashcode method, but that method
-            //is just returning the int value, I think, so
-            //this should potentially just be replaced by
-            //an int.
-            Integer mult = DataUtil.integer(getMultiplicity(i));
-            if (i == 0 && mult.intValue() == INDEX_UNBOUND)
-                mult = DataUtil.integer(0);
```
phillipm: use of `DataUtil.integer` here and above while assigning `hash` were completely unnecessary (see the NOTE comment)

core/src/main/java/org/javarosa/core/model/instance/TreeReferenceLevel.java
```diff
@@ -141,8 +141,8 @@ public boolean equals(Object o) {
         }
 
         // predicates match-up
-        if ((predicates == null && l.predicates != null) ||
-                (l.predicates == null && predicates != null) ||
+        if (predicates == null ||
+                l.predicates == null ||
```
phillipm: Remove check to ensure both predicates aren't `null`, since that is done right above on line 139.

core/src/main/java/org/javarosa/core/model/instance/TreeReferenceLevel.java
```diff
@@ -132,7 +132,7 @@ public boolean equals(Object o) {
         // multiplicity and names match-up
         if ((multiplicity != l.multiplicity) ||
                 (name == null && l.name != null) ||
-                (!name.equals(l.name))) {
+                (name != null && (!name.equals(l.name)))) {
```
phillipm: Add additional check to prevent NPE

PR #172, opened 2015-10-08
Add type annotation to vectors | hopefully it doesn't break J2ME builds

phillipm:
Also, remove unused `ActionWrapper`, make a few methods private, and make `Action` abstract to force implementation of `processAction`.


amstone326:
Hey, a little confused-- how would this not break j2me builds?

phillipm:
There are similar type annotations already existing in the javarosa repo. The J2ME pre-compilation step that removes generics seems pretty finicky, it only works in similar situations, for instance, the build server just crashed because of this line: https://github.com/dimagi/commcare/commit/d58e75520c68e80fada76b25768ac06181408008#diff-ea530c7fb1c9a43cc9fefc4942283ca9L427

Regardless, I just tested this branch on a jenkins j2me build and it compiled

PR #173, opened 2015-10-09
Nested Repeat Bugfix

ctsims:
Fix a bug with nested repeat actions firing for both the exterior repeat as well as the templated internal repeat. Essentially with the structure
```
<outer>
   <inner>
         <value>
```
where both outer and inner are repeats, an action which should fire when inner is created was being triggered on outer instead.


core/src/main/java/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -71,37 +71,70 @@ public void processAction(FormDef model, TreeReference contextRef) {
...
+                        }
+                    }
+                }
+
+                if (singleFullyQualifiedRef == null) {
+                    //We only have template references, which means this setvalue action isn't relevant
+                    //for these targets yet.
+                    return;
+                }
+            } else {
```
phillipm: This template checking logic is pretty self-contained and should be a helper function.

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -202,4 +207,35 @@ public void testSetValuePredicate() {
...
+                continue;
+            }
+        } while (fec.stepToNextEvent() != fec.EVENT_END_OF_FORM);
+
+        if(!xpathEvalAndCompare(fpi.getFormDef().getEvaluationContext(), "/data/sum", 30.0)) {
+            fail("Nested repeats did not evaluate to the proper outcome");
+        }
+    }
+
+    public static boolean xpathEvalAndCompare(EvaluationContext evalContext,
```
phillipm: Please move this into `ExprEvalUtils`

core/src/test/resources/xform_tests/test_looped_model_iteration.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<h:html xmlns:h="http://www.w3.org/1999/xhtml" xmlns:orx="http://openrosa.org/jr/xforms" xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:jr="http://openrosa.org/javarosa" xmlns:vellum="http://commcarehq.org/xforms/vellum">
+    <h:head>
```
phillipm: Would be nice to run the AS code formatter on this file. It makes it a bit more readable

core/src/main/java/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -71,37 +71,70 @@ public void processAction(FormDef model, TreeReference contextRef) {
...
-            else if(references.size() > 1) {
-                //The spec for single value binding says we should pick the first one,
-                //but that's not how we do things to prevent errors. We're going to fail
-                //the same way a nodeset does
-                throw new XPathTypeMismatchException("XPath nodeset has more than one node [" + XPathNodeset.printNodeContents(references) + "]; Actions can only target a single node reference. Refine path expression to match only one node.");
-            }
-            else {
+            } else if (references.size() > 1) {
+                //We included templates, so it's possible we got both a ref _and_ a template. Check
+                TreeReference singleFullyQualifiedRef = null;
```
phillipm: The test suite doesn't ever enter this conditional branch. Could you write a test to ensure the correctness of this newly added code?

ctsims:
@phillipm updated with your PR feedback and removed the logic that you were correct was not being used.

core/src/main/java/org/javarosa/core/model/actions/SetValueAction.java
```diff
@@ -68,37 +68,56 @@ public void processAction(FormDef model, TreeReference contextRef) {
...
+            } else {
                 qualifiedReference = references.elementAt(0);
             }
         }
 
+        //If after finding our concrete reference it is a template, this action is outside of the
+        //scope of the current target, so we can leave.
+        if(qualifiedReference.isTemplateRef()) {
+            return;
+        }
```
phillipm: This conditional is never satisfied by any tests in the test suite. Since this is in the depths of some pretty complex and uses the newly introduced `isTemplateRef`, I think it is important to write a test case that triggers a `return` here.

ctsims: Hm, that's a good point. I'll remove it, since I'm not actually sure how we could end up here, based on how expandReference actually ends up working.

ctsims:
@amstone326 Just addressed phillip's last comment. Can you pull this? it's blocking Matt's ICDS development

PR #174, opened 2015-10-09
Don't fire triggerables twice in repeat entry init

phillipm:
Say we are inserting an item into the `/data/country/` repeat, there are 3 trigger firing phases for this insertion:

1. Fire triggers tied to jr-insert events: `<setvalue event="jr-insert" ref="/data/country/@id" value="int(count(/data/country))"/>`
2. Fire triggerables that depend on the repeat: any triggerables that reference `/data/country`, which when evaluated in the current context, will resolve to the new entry we are inserting. For example, `/data/country[5]` if we are inserting the 5th element. These firings will cascade, so even if only one calculation depends on `/data/country`, many triggers may get fired if that calculation is used extensively in the form. 
3. Fire triggerables associated with sub elements of the repeat entry being inserted. When we insert a new country, we need to fire triggerables for `/data/country/language`, `/data/country/population`, etc.

It turns out that triggerables fired in _step 1_ don't need to be explicitly fired again in _step 3_. This PR ensures this doesn't happen. Of course, it may be that another triggerable fired in _step 3_ will cause that triggerable to need to be recalculated. The test added ensures that this is in fact occurring.

Reduces model iteration time when loading [this ICDS form](https://india.commcarehq.org/a/icds-test/apps/view/3c38e2d04aa3c9a2e2392da18743b1b9/modules-7/forms-0/source/) from 9 to 6 seconds. For form loading.

Also reduces model iteration for a TDH app by 1 second.


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -854,6 +876,35 @@ private void initializeTriggerables(TreeReference rootRef) {
...
+
+        Vector<Triggerable> applicable = new Vector<Triggerable>();
+        for (int i = 0; i < triggerables.size(); i++) {
+            Triggerable t = (Triggerable)triggerables.elementAt(i);
+            for (int j = 0; j < t.getTargets().size(); j++) {
+                TreeReference target = (TreeReference)t.getTargets().elementAt(j);
+                if (genericRoot.isParentOf(target, false)) {
+                    if (!triggerablesProcessedInSetAction.contains(t)) {
+                        applicable.addElement(t);
+                    }
```
phillipm: Doesn't add triggerable for evaluation if it is was processed by the insert action stage.

ctsims: Hey,

So one thing we need to make sure is taken into account here is if a triggerable has _two_ triggers, one of which is the setvalue and one of which is independent, we still want to trigger if the second one triggered this.

So basically:

setValue: A = 10
Calculate: C = 5
Calculate: B = A + C

the last calculate will fire during the setvalue action, but should _also_ fire because C was set during the second phase.

phillipm:
@ctsims Spent some time working on this and I'm feeling pretty confident that it is a valid optimization.

core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -35,7 +35,7 @@
 /**
  * Representation of an xpath function expression.
  *
- * All of the built-in xpath functions are included here, as well as the xpath type conversion logic
+ * All of the built-in xpath functions are included here, as wellaas the xpath type conversion logic
```
ctsims: typo I think

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -138,10 +144,13 @@
     private final CacheTable<TreeReference, Vector<TreeReference>> cachedCascadingChildren =
             new CacheTable<TreeReference, Vector<TreeReference>>();
 
-
     /**
-     *
+     * Store triggerables that were fired while processing an insert action for
+     * a newly created repeat entry.
      */
+    private Vector<Triggerable> triggerablesProcessedInSetAction =
```
ctsims: This temporary data set and the flag above are fairly coupled. 

Am wondering whether it would be worth it to declare and synchronize on a lock which protects the scoping in which these two values are being used. It would make things much more clear and slightly less error prone. In general it seems good to be protecting the "virtual scope" that is the set of methods that might be active with these members.

phillipm:
@ctsims `synchronized` always confuses me so I want to avoid using it. I refactored the PR to explicitly thread the list of fired triggerables. Can you take a look at it again?

core/src/main/java/org/javarosa/core/model/Action.java
```diff
@@ -33,8 +33,11 @@ public Action(String name) {
      * NOTE: Currently actions are only processed on nodes that are
      * WITHIN the context provided, if one is provided. This will
      * need to get changed possibly for future action types.
+     *
+     * @return TreeReference targeted by the action or null if the action
+     * wasn't completed.
      */
-    public abstract void processAction(FormDef model, TreeReference context);
+    public abstract TreeReference processAction(FormDef model, TreeReference context);
```
phillipm: Does changing an abstract method require any intervention with the J2ME builds?

PR #175, opened 2015-10-09
Cache cascading calc in fillTriggeredElements

phillipm:
Cache computation that finds generic references to children of a triggerable target. Makes use of templates when available, and shortcuts if the target is an attribute. Doesn't cache if the target reference points to a heterogeneous ref like `/data/a/b/d` found here: https://github.com/dimagi/javarosa/blob/master/core/src/test/resources/test_nested_preds_with_rel_refs.xml#L20-L30 . This would require an update to how templates are processed that I am weary of doing.

Evaluation is twice as fast when swiping forward past the first question of ICDS's _Child Due List_ form in the due list test app.


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -945,7 +960,8 @@ private void evaluateTriggerable(Triggerable t, TreeReference anchorRef) {
      * when we support dependent XPath Steps (IE: /path/to//)
      */
     public void addChildrenOfReference(TreeReference original, Vector<TreeReference> toAdd) {
-        for (TreeReference ref : exprEvalContext.expandReference(original)) {
+        Vector<TreeReference> refs = exprEvalContext.expandReference(original);
```
ctsims: Question: Are these references genericized when this gets fired? IE: Are the treerefrences

/data/path/to/node

or 

/data[1]/path[1]/to[3]/node[1]

phillipm: `original` is generic and `refs` is the list of possible nodes it can point to. Why? Hmmm, is it that I should be caching only after expanding, not before?

phillipm: Rather, what assumptions about structure homogeneity can we assume at this point? If we can assume that the target of triggers will be homogeneous, we can just cache the template and be done with this child traversal funny business (notice that we store generic refs to all the nodes collected this way, so we end up with duplicate references!)

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -851,7 +892,7 @@ private void initializeTriggerables(TreeReference rootRef) {
             }
         }
 
-        evaluateTriggerables(applicable, rootRef);
+        evaluateTriggerables(applicable, rootRef, true);
```
ctsims: This method gets called in contexts which aren't repeats, so either the name of this true arg (isRepeatEntryInit) isn't correct, or this value isn't

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -702,49 +709,83 @@ public void finalizeTriggerables() throws IllegalStateException {
...
-                //For certain types of triggerables, the update will affect not only the target, but
-                //also the children of the target. In that case, we want to add all of those nodes
-                //to the list of updated elements as well. For instances, relevancy of a parent will
-                // require triggers pointing to children to need to be recalcualted
-                if (t.isCascadingToChildren()) {
-                    addChildrenOfReference(target, updatedNodes);
+            for (TreeReference target : t.getTargets()) {
+                Vector<TreeReference> updatedNodes;
+                if (isRepeatEntryInit) {
+                    // children have already been added to 'destination'
```
ctsims: I'm not really sure I understand this, how is destination bootstrapped here? Just the fact that we're initializing ~every triggerable anyway?

ctsims: If the above, bear in mind that there could be references to the triggerable conditions _outside_ of the repeat, which wouldn't be triggered based on this.,

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -528,7 +540,7 @@ public void copyItemsetAnswer(QuestionDef q, TreeElement targetNode, IAnswerData
         triggerTriggerables(destRef);
 
         // initialize conditions for the node (and sub-nodes)
-        initializeTriggerables(destRef);
+        initTriggerablesRootedBy(destRef);
```
phillipm: @ctsims Can you give context on when this method (`copyItemsetAnswer`) is used? It is only called for complex questions in `FormEntryController.anserQuestion`. Supposedly complex questions don't really occur often? Is it okay that this call here is calling `fillTriggeredElements` with `isRepeatEntryInit` set to true? or should I isolate it?

core/src/main/java/org/javarosa/core/model/FormDef.java
phillipm: Fixed this, other than the comment I left above.

core/src/main/java/org/javarosa/core/model/instance/DataInstance.java
```diff
@@ -125,50 +125,6 @@ public T resolveReference(TreeReference ref) {
         return t;
     }
 
-    public Vector explodeReference(TreeReference ref) {
```
phillipm: Dead code

core/src/main/java/org/javarosa/core/model/instance/XmlWalkingTreeElement.java
```diff
@@ -1,255 +0,0 @@
...
-import java.util.Vector;
-
-/**
- * The root element for the <casedb> abstract type. All children are
- * nodes in the case database. Depending on instantiation, the <casedb>
- * may include only a subset of the full db.
- *
- * @author ctsims
- */
-public class XmlWalkingTreeElement implements AbstractTreeElement<XmlWalkingTreeElement> {
```
phillipm: Unused class

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -2397,23 +2396,6 @@ private static void trimRepeatChildren(TreeElement node) {
         }
     }
 
-    private static void checkDuplicateNodesAreRepeatable(TreeElement node) {
```
phillipm: Dead code

phillipm:
This is ready to review again, though I'd like to do another pass to add comments, rename some methods, and make sure the tests are concise and functional

core/src/main/java/org/javarosa/core/model/FormDef.java
ctsims: Hey,

CopyItemsetAnswer is completely unused but was included for spec compliance. When you use it, it copies the full subtree of the itemset nodeset for the selected choices. For now this is almost certainly fine, I'd probably make a note that it should possibly exclude subtrees copied over in that way, though.

ctsims:
@phillipm This looks like a totally reasonable approach and a good optimization. I think it's good for cleanup/final push

PR #185, opened 2015-10-26
Don't return false if we can't find template

wpride:
@ctsims @phillipm possible fix for http://manage.dimagi.com/default.asp?181597

Don't think I completely understand when we want to use template vs. when we want to fall back to normal nodes, though, so glad for feedback


phillipm:
Can this be closed now that https://github.com/dimagi/touchforms/pull/176 has been merged?

wpride:
@phillipm yeah this isn't needed anymore, thanks

PR #188, opened 2015-10-29
Dynamic refreshing of question content in question lists

amstone326:
Previously, if the question text or select options for one question in a question list were dependent upon the answer to a prior question in the group, the question text / select options would NOT get updated when a user changed their answer to the prior question. This fixes that.

cross: https://github.com/dimagi/commcare-odk/pull/762

### DIFF
--- a/core/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
+++ b/core/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
@@ -235,17 +235,14 @@ public class FormEntryPrompt extends FormEntryCaption {
         }
     }
 
-    private boolean questionTextIsUnchanged(FormEntryPrompt oldPrompt) {
-        String newQuestionText = getQuestionText();
-        String oldQuestionText = oldPrompt.getQuestionText();
-        return newQuestionText.equals(oldQuestionText);
+    private boolean questionTextIsUnchanged(String oldQuestionText) {
+        return getQuestionText().equals(oldQuestionText);
     }
 
-    public boolean hasSameDisplayContent(FormEntryPrompt oldPrompt,
+    public boolean hasSameDisplayContent(String questionTextForOldPrompt,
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -206,21 +206,55 @@ public String getConstraintText(String textForm, IAnswerData attemptedValue) {
...
+        } else {
+            return newQuestionText.equals(oldQuestionText);
+        }
+    }
+
+    /**
+     * @return If this prompt has all of the same display content as a previous prompt that had
+     * the given question text and select choices
+     */
+    public boolean hasSameDisplayContent(String questionTextForOldPrompt,
```
wpride: I don't see where this method is ever called? And the two methods above are only called by this method

amstone326: This is cross requested, look at the other PR.

wpride: classic

core/src/main/java/org/javarosa/form/api/FormEntryPrompt.java
```diff
@@ -206,21 +206,55 @@ public String getConstraintText(String textForm, IAnswerData attemptedValue) {
...
+    private boolean selectChoicesAreUnchanged(Vector<SelectChoice> choicesForOld) {
+        Vector<SelectChoice> choicesForThis = getSelectChoices();
+        if (choicesForOld == null) {
+            return choicesForThis == null;
+        } else {
+            return choicesForOld.equals(choicesForThis);
+        }
+    }
+
+    private boolean questionTextIsUnchanged(String oldQuestionText) {
```
phillipm: move definition below usage in class

ctsims:
@amstone326 this looks fine to me. Left a question in the android PR and can pull afterwards

PR #191, opened 2015-11-03
Catch out-of-bounds exception in selected-at

phillipm:
Throw the right exception when processing a xpath with `select-at` that attempts to select outside of the range.

Fix for crashes like https://commcare-acralyzer.cloudant.com/acralyzer/_design/acralyzer/index.html#/reports-browser/commcare-odk/bug/88e06e6f6fefd6e46c48cc88461f88d5

### DIFF
--- a/core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -865,7 +865,7 @@ public class XPathFuncExpr extends XPathExpression {
 
         Vector<String> entries = DateUtils.split(selection, " ", true);
 
-        if (entries.size() < index) {
+        if (entries.size() <= index) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -863,7 +863,14 @@ public static String selectedAt(Object o1, Object o2) {
...
 
-        return DateUtils.split(selection, " ", true).elementAt(index);
+        Vector<String> entries = DateUtils.split(selection, " ", true);
+
+        if (entries.size() <= index) {
+            throw new XPathException("Attempting to select element " + index +
+                    " of a list with only " + entries.size() + " elements.");
+        } else {
+            return entries.elementAt(index);
+        }
```
phillipm: Fix; the rest are tests and moving test util methods to util files

PR #195, opened 2015-11-13
Check for DateTimeData in unpackValue

phillipm:
Don't really understand the implications of this change, but Sheel was getting `warning: unrecognized data type in xpath expr: org.javarosa.core.model.data.DateTimeData` and this fixes that.

Might be a fix for http://manage.dimagi.com/default.asp?186983


core/src/main/java/org/javarosa/xpath/expr/XPathPathExpr.java
```diff
@@ -254,6 +255,8 @@ public static Object unpackValue(IAnswerData val) {
             return (new XFormAnswerDataSerializer()).serializeAnswerData(val);
         } else if (val instanceof DateData) {
             return val.getValue();
+        } else if (val instanceof DateTimeData) {
+            return val.getValue();
```
phillipm: the fix

PR #197, opened 2015-11-30
Refactor: Linting around logging code

phillipm:


core/src/main/java/org/javarosa/core/services/Logger.java
```diff
@@ -110,10 +110,6 @@ public void run() {
         throw crashException;
     }
 
-    public static void crashTest(String msg) {
-        throw new FatalException(msg != null ? msg : "shit has hit the fan");
```
orangejenny: thanks, Drew

orangejenny: (i assume)

PR #198, opened 2015-12-02
Contextualization was conflating attributes and normal nodes of the same name

phillipm:
When we contextualize the `current()` in something like `<bind nodeset="/data/treatments/iterate_over_meds/item/id" calculate="current()/../@id"/>` we conflate `@id` and `/id`, such that `@id` gets set instead of `/id`.

This wasn't coming up because we were always re-triggering thing twice which somehow hid this issue. But since https://github.com/dimagi/javarosa/pull/174 triggers are only fired once, so this issue arises.

The test cases I added fail when the fix is absent.

http://manage.dimagi.com/default.asp?188205

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -988,8 +988,8 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
                 if (genericRoot.isParentOf(target, false)) {
                     if (!triggeredDuringInsert.contains(triggerable)) {
                         applicable.addElement(triggerable);
+                        break;
                     }
-                    break;
                 }
             }
         }
--- a/core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
+++ b/core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
@@ -452,8 +452,14 @@ public class TreeReference implements Externalizable {
             if (contextRef.getName(i).equals(newRef.getName(i))) {
                 // Only copy over multiplicity info if it won't overwrite any
                 // existing preds or filters
+
+                // don't copy multiplicity from context when new ref's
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -452,8 +452,14 @@ public TreeReference contextualize(TreeReference contextRef) {
...
                 // existing preds or filters
+
+                // don't copy multiplicity from context when new ref's
+                // multiplicity is already bound or when the context's
+                // multiplicity is not a position (but rather an attr or
+                // template)
                 if (newRef.getPredicate(i) == null &&
-                        newRef.getMultiplicity(i) == INDEX_UNBOUND) {
+                        newRef.getMultiplicity(i) == INDEX_UNBOUND &&
+                        contextRef.getMultiplicity(i) >= 0) {
```
phillipm: @ctsims not copying over multiplicities below 0 from the context ref seems like the right call here. The call to `anchor` up above should grab those negative multiplicities, so attrs should still be setup correctly.

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -988,8 +988,8 @@ private void initTriggerablesRootedBy(TreeReference rootRef,
                 if (genericRoot.isParentOf(target, false)) {
                     if (!triggeredDuringInsert.contains(triggerable)) {
                         applicable.addElement(triggerable);
+                        break;
```
phillipm: Move `break` such that it doesn't after iterating through the first target. We want to continue looping through the triggerable's targets to make sure that if one of them wasn't already processed during the repeat insert phase then we add it to process now.

Not entirely sure it is the right thing to do, but it will cast a larger net so it doesn't matter if it isn't

PR #202, opened 2015-12-03
Serialize attributes correctly

phillipm:
Serializing TreeElement's attributes looses the data type encoded in it, causing problems with data storage, like xforrm dateTime types being treated as dates.

Attribute serialization has always been broken but I think some redundancy in trigger firing has masked that fact. The issue has probably arisen now that I've optimized trigger firing.

http://manage.dimagi.com/default.asp?188492


phillipm:
Re-opened to target 2.25 https://github.com/dimagi/javarosa/pull/204

PR #205, opened 2015-12-08
use non main instance

wpride:


wpride:
see https://github.com/dimagi/commcare/pull/218

phillipm:
Should target 2.25 if it is cross-requested with https://github.com/dimagi/commcare/pull/218

phillipm:
Recreated at https://github.com/dimagi/javarosa/pull/206

PR #206, opened 2015-12-08
use non main instance

wpride:


wpride:
cross https://github.com/dimagi/commcare/pull/218

PR #207, opened 2015-12-08
Fix for Navigation Bug with End-Form Repeats

ctsims:
Context: http://manage.dimagi.com/default.asp?188096#1057212

Fix for bug where repeats were attempting to be contextualized against an incomplete reference when their creation was forced due to navigation. Basically we were using the "local" portion of a form index's reference, the piece for the current 'step' instead of representing the full index.

### DIFF
--- a/core/src/main/java/org/javarosa/form/api/FormEntryModel.java
+++ b/core/src/main/java/org/javarosa/form/api/FormEntryModel.java
@@ -426,7 +426,7 @@ public class FormEntryModel {
             if (e instanceof GroupDef) {
                 GroupDef g = (GroupDef)e;
                 if (g.getRepeat() && g.getCountReference() != null) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/javarosa/form/api/FormEntryModel.java
```diff
@@ -426,13 +426,13 @@ public void createModelIfNecessary(FormIndex index) {
             if (e instanceof GroupDef) {
                 GroupDef g = (GroupDef)e;
                 if (g.getRepeat() && g.getCountReference() != null) {
-                    IAnswerData count = getForm().getMainInstance().resolveReference(g.getConextualizedCountReference(index.getLocalReference())).getValue();
+                    IAnswerData count = getForm().getMainInstance().resolveReference(g.getConextualizedCountReference(index.getReference())).getValue();
```
ctsims: This is the fix

PR #209, opened 2015-12-09
Store fixtures in files instead of DB

phillipm:
Cross-requested with https://github.com/dimagi/commcare-odk/pull/906 and https://github.com/dimagi/commcare/pull/221


core/src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -7,10 +7,24 @@
...
-    private StreamsUtil() {
-        // private constructor
+    public static byte[] getStreamAsBytes(InputStream is) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+        try {
+            StreamsUtil.writeFromInputToOutput(is, baos);
+            return baos.toByteArray();
+        } catch (IOException e) {
+            e.printStackTrace();
```
ctsims: this seems wrong to me v. just throwing an IOException as part of the method signature is there a reason we went with this?

core/src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -45,9 +59,7 @@ public static void writeFromInputToOutputSpecific(InputStream in, OutputStream o
     public static void writeFromInputToOutput(InputStream in, OutputStream out) throws IOException {
         try {
             writeFromInputToOutput(in, out, null);
-        } catch (InputIOException e) {
-            throw e.internal;
-        } catch (OutputIOException e) {
+        } catch (OutputIOException | InputIOException e) {
```
ctsims: Not Java 1.3 compatible

core/src/main/java/org/javarosa/core/io/StreamsUtil.java
phillipm: Updated; see corresponding odk commit https://github.com/dimagi/commcare-odk/commit/0876d0cd3b88cab402efdd117f88a6774e960eb7

core/src/main/java/org/javarosa/core/io/StreamsUtil.java
```diff
@@ -7,10 +7,18 @@
...
-    private StreamsUtil() {
-        // private constructor
+    public static byte[] getStreamAsBytes(InputStream is) throws IOException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+
+        try {
+            StreamsUtil.writeFromInputToOutput(is, baos);
+            return baos.toByteArray();
+        } finally {
+            baos.close();
```
ctsims: minor, but same comment as in the other thread: do we know whether we think it's best practice to not swallow this here? We swallow closes in other contexts, but this might be a better way to do it.

phillipm: I once forgot to close something and it broke a lot of my logic. Based on that experience I think it is probably sorta important to crash if something fails to close. So I'd vote for leaving this how it is, but no especially well informed reason for this opinion

ctsims: That's fair, especially if we're buffering there is flushing behavior that can become relevant on close. We should probably keep an eye out for that kind of thing around the codebase in general.

PR #212, opened 2015-12-09
Validate cases DAG before case purge

amstone326:
Changes to DAG class to support: https://github.com/dimagi/commcare/pull/223
(also cross: https://github.com/dimagi/commcare-odk/pull/910)


core/src/main/java/org/javarosa/core/util/DAG.java
```diff
@@ -19,51 +19,90 @@
...
     }
 
     public void addNode(I i, N n) {
         nodes.put(i, n);
     }
 
+    public N removeNode(I i) {
+        N removedNode = getNode(i);
+        nodes.remove(i);
+        return removedNode;
```
phillipm: This method can be simplified to `return nodes.remove(i);`

amstone326: yikes, true.

PR #219, opened 2016-01-08
Geo distance function

forest-tong:


core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1431,6 +1437,21 @@ public static double pow(final double a, final double b) {
     public static final double DOUBLE_TOLERANCE = 1.0e-12;
 
     /**
+     * Returns the distance between two GeoPointData locations, in meters, given objects to unpack.
+     * Ignores altitude and accuracy.
+     */
+    public static double distance(Object from, Object to) {
+        String unpacked_from = (String) unpack(from);
```
phillipm: To follow Java naming conventions: `unpack_from` -> `unpackFrom`

core/src/test/java/org/javarosa/core/model/test/GeoPointTests.java
```diff
@@ -0,0 +1,52 @@
...
+    @Before
+    public void setUp() {
+        geopointFpi = new FormParseInit("/geopoint_tests.xml");
+
+        FormEntryController fec = geopointFpi.getFormEntryController();
+        fec.jumpToIndex(FormIndex.createBeginningOfFormIndex());
+
+        // step through entire form
+        do {
+        } while (fec.stepToNextEvent() != FormEntryController.EVENT_END_OF_FORM);
```
phillipm: You can actually remove lines 21 - 26 since none of the tests actually make use of more complex form logic.

phillipm: (I had left that in the example I sent you in case we wanted to go there, but doesn't seem necessary at this point)

core/src/test/java/org/javarosa/core/model/test/GeoPointTests.java
```diff
@@ -0,0 +1,52 @@
...
+                geopointEvalCtx);
+    }
+
+    /**
+     * Tests distance from New York to San Francisco with a tolerance of 10km
+     */
+    @Test
+    public void testDistanceFunctionBetweenDifferentPoints() throws XPathSyntaxException {
+        ExprEvalUtils.assertAlmostEqualsXpathEval(
+                4128850.0, 10000.0, "distance(/data/new_york, /data/san_francisco)", geopointEvalCtx);
```
phillipm: What is the reason for a 10km tolerance? Is it just to suggest that these calculations are done over uncertain data or do the results actually contain variation between invocations?

phillipm:
Awesome! Left some small actionable comments and a question. Will merge after the comments are addressed

core/src/test/java/org/javarosa/core/model/test/GeoPointTests.java
forest-tong: The result doesn't vary between invocations, I just built in the tolerance because double calculations are fickle (in case the distance calculation is refactored or something, for example). Maybe 10km is too much, so I refined the number and changed it to 1m, which is still somewhat arbitrary of a tolerance.

core/src/test/java/org/javarosa/core/model/test/GeoPointTests.java
forest-tong: Done.

PR #230, opened 2016-01-15
Clean up User.java during PIN password work

amstone326:
cross: https://github.com/dimagi/commcare-odk/pull/979, https://github.com/dimagi/commcare/pull/243


core/src/main/java/org/javarosa/core/model/User.java
```diff
@@ -102,8 +102,8 @@ public String getUsername() {
         return username;
     }
 
-    public String getPassword() {
-        return password;
+    public String getPasswordHash() {
```
ctsims: @amstone326 I think we use this in the CommCare j2me builds as well. Worth taking a look to see if it needs to change outside of /core/ (which you already took a look at)

amstone326: Good call, thanks. fixing.

PR #233, opened 2016-01-20
Distance between empty locations

forest-tong:
Specification of distance function changed to return -1 when either argument is an empty string (or null). 

Previously, an exception would be thrown if an argument was null, and the location (0, 0, 0, 0) would be used for the empty string.

### DIFF
--- a/core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
+++ b/core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
@@ -1466,7 +1466,7 @@ public class XPathFuncExpr extends XPathExpression {
         String unpackedFrom = (String) unpack(from);
         String unpackedTo = (String) unpack(to);
 
-        if (unpackedFrom == "" || unpackedFrom == null || unpackedTo == "" || unpackedTo == null) {
+        if (unpackedFrom == null || "".equals(unpackedFrom) || unpackedTo == null || "".equals(unpackedTo)) {
             return -1.0;
         }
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1460,11 +1460,16 @@ public static double pow(final double a, final double b) {
      * Returns the distance between two GeoPointData locations, in meters, given objects to unpack.
      * Ignores altitude and accuracy.
      * Note that the arguments can be strings.
+     * Returns -1 if one of the arguments is null or the empty string.
      */
     public static Double distance(Object from, Object to) {
         String unpackedFrom = (String) unpack(from);
         String unpackedTo = (String) unpack(to);
 
+        if (unpackedFrom == "" || unpackedFrom == null || unpackedTo == "" || unpackedTo == null) {
```
phillipm: `someString == ""` does a reference level equality check in Java, which isn't what you want here. 
Should be `someString == null || "".equals(someString)`

core/src/main/java/org/javarosa/xpath/expr/XPathFuncExpr.java
```diff
@@ -1460,11 +1460,16 @@ public static double pow(final double a, final double b) {
...
      * Ignores altitude and accuracy.
      * Note that the arguments can be strings.
+     * Returns -1 if one of the arguments is null or the empty string.
      */
     public static Double distance(Object from, Object to) {
         String unpackedFrom = (String) unpack(from);
         String unpackedTo = (String) unpack(to);
 
+        if (unpackedFrom == null || "".equals(unpackedFrom) || unpackedTo == null || "".equals(unpackedTo)) {
+            return -1.0;
```
phillipm: You need to explicitly return a Double object of value -1, since J2ME doesn't auto-box. Build just :fire: :boom: :bomb: :boom: :fire:

PR #236, opened 2016-01-26
override SelectChoice .equals() method to produce correct functionali…

amstone326:
Fix for: http://manage.dimagi.com/default.asp?208770, which was introduced by https://github.com/dimagi/commcare-odk/pull/762. 

The issue as far as I can tell is this -- What is actually causing the user to be unable to select any options from the drop-down is that the code goes into an infinite loop when you view that question. The cause of that infinite loop is: FormEntryPrompt.selectChoicesAreUnchanged() checks for equality of the 2 SelectChoice vectors. In the case described in the ticket, selectChoicesAreUnchanged() is returning false, when it should be returning true, because .equals() on individual SelectChoices that are semantically equivalent is returning false. This is then causing a new form entry prompt to be added to the view when it really shouldn't be, and triggering the same code path to be called again.. etc. 

The ticket notes that the problem only comes up for single select *lookup tables*. When I was testing this change, I must not have tested it with any lookup table select widgets; just a normal select widget. My best guess is that the SelectChoices for a lookup table are re-created each time they are populated, meaning that now 2 different objects are being compared with .equals() and failing. 

tl;dr -- the solution (I think) is to override .equals() for SelectChoice. But I'm not sure if I'm doing it correctly, in terms of which fields should be included, and also if the corresponding hashCode() function is sufficient.

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/core/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -121,4 +121,46 @@ public class SelectChoice implements Externalizable, Localizable {
     public void setTextID(String textID) {
         this.textID = textID;
     }
+
+    @Override
+    public int hashCode() {
+        int result;
+        result = textID != null ? textID.hashCode() : 0;
+        result = 31 * result + value.hashCode();
+        result = 31 * result + index;
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+
+        if (!(obj instanceof SelectChoice)) {
+            return false;
+        }
+
+        SelectChoice otherChoice = (SelectChoice)obj;
+
+        String otherTextID = otherChoice.getTextID();
+        if (otherTextID == null) {
+            if (this.textID != null) {
+                return false;
+            }
+        }
+        else if (!otherChoice.getTextID().equals(this.textID)) {
+            return false;
+        }
+
+        if (!otherChoice.getValue().equals(this.value)) {
+            return false;
+        }
+
+        if (otherChoice.getIndex() != this.index) {
+            return false;
+        }
+
+        return true;
+    }
 }
--- a/core/src/main/java/org/javarosa/core/model/SelectChoice.java
+++ b/core/src/main/java/org/javarosa/core/model/SelectChoice.java
@@ -79,6 +79,9 @@ public class SelectChoice implements Externalizable, Localizable {
         return index;
     }
 
+    public boolean isLocalizable() {
+        return this.isLocalizable;
+    }
 
     public void localeChanged(String locale, Localizer localizer) {
 //        if (captionLocalizable) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -121,4 +121,46 @@ public String getTextID() {
     public void setTextID(String textID) {
         this.textID = textID;
     }
+
+    @Override
+    public int hashCode() {
+        int result;
+        result = textID != null ? textID.hashCode() : 0;
+        result = 31 * result + value.hashCode();
```
phillipm: Looks like `value` can be `null`

core/src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -121,4 +121,46 @@ public String getTextID() {
...
+
+        SelectChoice otherChoice = (SelectChoice)obj;
+
+        String otherTextID = otherChoice.getTextID();
+        if (otherTextID == null) {
+            if (this.textID != null) {
+                return false;
+            }
+        }
+        else if (!otherChoice.getTextID().equals(this.textID)) {
```
phillipm: `otherChoice.getTextID()` -> `otherTextID`

core/src/main/java/org/javarosa/core/model/SelectChoice.java
```diff
@@ -121,4 +121,46 @@ public String getTextID() {
     public void setTextID(String textID) {
         this.textID = textID;
     }
+
+    @Override
+    public int hashCode() {
+        int result;
+        result = textID != null ? textID.hashCode() : 0;
+        result = 31 * result + value.hashCode();
+        result = 31 * result + index;
```
phillipm: I think usually `hashCode` implementations use xor (`^`). Is there a reason for the `31 *`?

amstone326: Yea, was honestly just copying the structure of sample implementations. Think you're right about xor though, can add that.

amstone326:
@ctsims Can you weight in on what semantic equality of SelectChoices would be?

phillipm:
My take, not being familiar with `SelectChoice`, is that it is probably a good idea to also check values of `isLocalizable` and `labelInnerText` in `hashCode` and `equals` implementations. Probably doesn't hurt to leave it out though...

phillipm:
Worth putting into 2.26?

amstone326:
Definitely. I'm actually kind of unclear on why this issue didn't come up for a lot of users, since I don't think of single-select lookup tables as an uncommon thing. But should definitely get it in.

core/src/main/java/org/javarosa/core/model/SelectChoice.java
amstone326: The constructor throws an exception if value is null.

phillipm: ah, right. I was paying attention to when it is set via `ExtUtil.nullIfEmpty` 
https://github.com/dimagi/javarosa/blob/master/core/src/main/java/org/javarosa/core/model/SelectChoice.java#L93

core/src/main/java/org/javarosa/core/model/SelectChoice.java
amstone326: (I assume 31 is used because it's prime)

core/src/main/java/org/javarosa/core/model/SelectChoice.java
amstone326: @ctsims Not sure how to determine how much complexity is really necessary here. Some online examples I've found use xor, some don't seem to bother. Any thoughts?

PR #240, opened 2016-01-28
Add back in function used by touchforms

wpride:
Add back in a function that was removed here https://github.com/dimagi/javarosa/commit/6aedc0570c521a272b7ca0846764bcd2c52c9dba

Needed by touchforms

@phillipm

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1365,8 +1365,6 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
 
         }
 
-        System.out.println("121 populate choices: " + choices);
-
         itemset.setChoices(choices, this.getLocalizer());
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/javarosa/core/model/FormIndex.java
```diff
@@ -438,4 +438,28 @@ public static boolean overlappingLocalIndexesMatch(FormIndex parent, FormIndex c
         }
         return parent.getLocalIndex() == child.getLocalIndex();
     }
+
+    public void assignRefs(FormDef f) {
```
phillipm: Can you mention that this is needed by touchforms so I don't remove it in the future?

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1365,6 +1365,8 @@ public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef
 
         }
 
+        System.out.println("121 populate choices: " + choices);
```
phillipm: Would be nice to not include this debug print

phillipm:
For touchforms jar deployment, do you want this in 2.26 or is it fine just being on master?

wpride:
doesn't particularly matter

PR #243, opened 2016-02-03
Question level actions support

amstone326:
Add support to the xforms engine for actions at a per-question level. Will require a bunch of HQ changes to be useful (though can be merged without them since it will just have no effect). Draft of spec here: https://docs.google.com/a/dimagi.com/document/d/1Llx1xf1MNCNJo5VjxgQNvZy_PioYbc1raIuSeA_y70E/edit?usp=sharing

cross: https://github.com/dimagi/commcare-odk/pull/1047


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -372,45 +373,32 @@ public FormIndex deleteRepeat(FormIndex index) {
...
+        mainInstance.copyNode(template, repeatContextRef);
 
-        preloadInstance(mainInstance.resolveReference(destRef));
+        preloadInstance(mainInstance.resolveReference(repeatContextRef));
 
         // Fire jr-insert events before "calculate"s
-        Vector<Triggerable> triggeredDuringInsert = new Vector<Triggerable>();
-        processInsertAction(destRef, triggeredDuringInsert);
+        triggeredDuringInsert = new Vector<Triggerable>();
+        triggerActionsFromEvent(Action.EVENT_JR_INSERT, this, repeatContextRef, this);
```
amstone326: Refactored to allow EVENT_JR_INSERT to utilize the existing framework, instead of calling its own custom `processInsertAction` method

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1859,30 +1848,7 @@ public SubmissionProfile getSubmissionProfile() {
         return submissionProfiles.get(DEFAULT_SUBMISSION_PROFILE);
     }
 
-    public Vector<Action> getEventListeners(String event) {
```
amstone326: moved to `ActionController` abstract class, which this class now extends

core/src/main/java/org/javarosa/core/model/QuestionDef.java
```diff
@@ -153,12 +154,6 @@ public void removeSelectChoice(SelectChoice choice) {
         }
     }
 
-    public void removeAllSelectChoices() {
```
amstone326: unused

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -124,8 +125,7 @@
     private String defaultNamespace;
     private Vector<String> itextKnownForms;
 
-    private static Vector<String> namedActions;
```
amstone326: unused

core/src/main/java/org/javarosa/core/model/actions/Action.java
```diff
@@ -48,4 +44,22 @@ public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOExcep
...
+
+    public static boolean isValidEvent(String s)  {
+        String[] allEvents = new String[] {EVENT_JR_INSERT,
+                EVENT_QUESTION_VALUE_CHANGED, EVENT_XFORMS_READY, EVENT_XFORMS_REVALIDATE};
+        for (String event : allEvents) {
+            if (event.equals(s)) {
+                return true;
+            }
+        }
+        return false;
```
phillipm: I would make `allEvents` a `private static final Vector<String>` so that it is only constructed one time, not every time `isValidEvent` is called.
Then this can be `isValidEvent(String actionEventAttr) { return allEvents.contains(actionEventAttr); }`

core/src/main/java/org/javarosa/core/model/actions/Action.java
```diff
@@ -48,4 +44,22 @@ public void readExternal(DataInputStream in, PrototypeFactory pf) throws IOExcep
     public void writeExternal(DataOutputStream out) throws IOException {
         ExtUtil.writeString(out, name);
     }
+
+    // Events that can trigger an action
+    public static final String EVENT_XFORMS_READY = "xforms-ready";
+    public static final String EVENT_XFORMS_REVALIDATE = "xforms-revalidate";
+    public static final String EVENT_JR_INSERT = "jr-insert";
+    public static final String EVENT_QUESTION_VALUE_CHANGED = "xforms-value-changed";
```
phillipm: These should be defined at the top of the class defintion.

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -126,10 +128,10 @@
 
     private FormInstance mainInstance = null;
 
-    Hashtable<String, Vector<Action>> eventListeners;
-
     boolean mDebugModeEnabled = false;
 
+    Vector<Triggerable> triggeredDuringInsert;
```
phillipm: `private` and consider making it `final` and then just clearing it on line 384.

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
```diff
@@ -0,0 +1,82 @@
...
+        if (eventListeners.containsKey(event)) {
+            actions = eventListeners.get(event);
+        } else {
+            actions = new Vector<Action>();
+            eventListeners.put(event, actions);
+        }
+        actions.addElement(action);
+    }
+
+    public void triggerActionsFromEvent(String event, FormDef model, TreeReference contextForAction) {
```
phillipm: All invocations of this method pass `null` in for `contextForAction`. Consider removing that arg.

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public void triggerActionsFromEvent(String event, FormDef model, TreeReference contextForAction,
+                                        ActionResultProcessor resultProcessor) {
+        for (Action action : getListenersForEvent(event)) {
+            TreeReference refSetByAction = action.processAction(model, contextForAction);
+            if (resultProcessor != null && refSetByAction != null) {
+                resultProcessor.processResultOfAction(refSetByAction, event);
+            }
+        }
+    }
```
phillipm: The method overloading here is a bit messy with the usage of `null` to toggle different behavior. I would recommend breaking them up into two methods: 
```
public void triggerActionsFromEvent(String event, FormDef model) {
    for (Action action : getListenersForEvent(event)) {
        action.processAction(model, contextForAction);
    }
}
```
and 
```
public void triggerActionsFromEventWithPostProcessing(String event, FormDef model, TreeReference contextForAction, ActionResultProcessor resultProcessor) {
    for (Action action : getListenersForEvent(event)) {
        TreeReference refSetByAction = action.processAction(model, contextForAction);
        resultProcessor.processResultOfAction(refSetByAction, event)
    }
}

```

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
```diff
@@ -0,0 +1,82 @@
...
+ * of those actions when an event occurs.
+ *
+ * @author Aliza Stone
+ */
+public abstract class ActionController implements Externalizable {
+
+    // map from an event to the actions it should trigger
+    private Hashtable<String, Vector<Action>> eventListeners = new Hashtable<String, Vector<org.javarosa.core.model.actions.Action>>();
+
+    public Vector<Action> getListenersForEvent(String event) {
```
phillipm: this can be `private`

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
```diff
@@ -0,0 +1,82 @@
...
+            TreeReference refSetByAction = action.processAction(model, contextForAction);
+            if (resultProcessor != null && refSetByAction != null) {
+                resultProcessor.processResultOfAction(refSetByAction, event);
+            }
+        }
+    }
+
+    @Override
+    public void readExternal(DataInputStream inStream, PrototypeFactory pf) throws IOException, DeserializationException {
+        eventListeners = (Hashtable<String, Vector<org.javarosa.core.model.actions.Action>>) ExtUtil.read(inStream,
```
phillipm: `org.javarosa.core.model.actions.Action` -> `Action`

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Registers actions that should be triggered by certain events, and handles the triggering
+ * of those actions when an event occurs.
+ *
+ * @author Aliza Stone
+ */
+public abstract class ActionController implements Externalizable {
+
+    // map from an event to the actions it should trigger
+    private Hashtable<String, Vector<Action>> eventListeners = new Hashtable<String, Vector<org.javarosa.core.model.actions.Action>>();
```
phillipm: `org.javarosa.core.model.actions.Action` -> `Action`

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
phillipm: oh, I missed the need of checking `refSetByAction != null`. My advice above is generally motivated by my dislike of `null` and method overloading. Feel free to ignore it if you don't agree

core/src/main/java/org/javarosa/form/api/FormEntryCaption.java
```diff
@@ -414,4 +414,8 @@ public void formElementStateChanged(TreeElement instanceNode,
     protected String getTextID() {
         return this.textID;
     }
+
+    public FormDef getForm() {
```
phillipm: doesn't seem to be used

core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1056,6 +1055,26 @@ protected QuestionDef parseControl(IFormElement parent, Element e, int controlTy
         return question;
     }
 
+    private void parseControlChildren(Element e, QuestionDef question, IFormElement parent,
+                                      boolean isSelect) {
+        for (int i = 0; i < e.getChildCount(); i++) {
+            int type = e.getType(i);
+            Element child = (type == Node.ELEMENT ? e.getElement(i) : null);
+            String childName = (child != null ? child.getName() : null);
```
phillipm: if `child == null` then we should be done here, all the subsequent logic expects `childName` to be something that isn't `null`

phillipm:
@ctsims @amstone326 this PR changes how `FormDef` is serialized, will this require a db migration?

core/src/main/java/org/javarosa/core/model/QuestionDef.java
```diff
@@ -48,7 +49,7 @@
  *
  * @author Daniel Kayiwa/Drew Roos
  */
-public class QuestionDef implements IFormElement, Localizable {
+public class QuestionDef extends ActionController implements IFormElement, Localizable {
```
phillipm: Can you point out where the fact that `QuestionDef` extends `ActionController` is utilized?

core/src/main/java/org/javarosa/core/model/QuestionDef.java
```diff
@@ -236,6 +231,7 @@ public IFormElement getChild(int i) {
      * @see org.javarosa.core.util.Externalizable#readExternal(java.io.DataInputStream)
      */
     public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOException, DeserializationException {
+        super.readExternal(dis, pf);
         setID(ExtUtil.readInt(dis));
         binding = (XPathReference)ExtUtil.read(dis, new ExtWrapNullable(new ExtWrapTagged()), pf);
         setAppearanceAttr((String) ExtUtil.read(dis, new ExtWrapNullable(String.class), pf));
```
phillipm: Serialization of `QuestionDef` isn't covered in the javarosa or commcare test suites. Can you look into how hard it would be to add test coverage of `QuestionDef`? This is a pretty straight forward change, but it would still be nice to cover it in the test suite, since serialization-related crashes are sometimes only caught in QA/after release.

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1469,6 +1457,7 @@ private boolean postProcessInstance(TreeElement node) {
      * @param dis - the stream to read from.
      */
     public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOException, DeserializationException {
+        super.readExternal(dis, pf);
         setID(ExtUtil.readInt(dis));
         setName(ExtUtil.nullIfEmpty(ExtUtil.readString(dis)));
         setTitle((String)ExtUtil.read(dis, new ExtWrapNullable(String.class), pf));
```
phillipm: Serialization of `FormDef` is also not covered by the test suites. Might be worth looking into adding it, if it isn't too difficult

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -247,6 +237,68 @@ public void testRepeatInsertTriggering() throws Exception {
...
+                new FormParseInit("/xform_tests/test_question_level_actions.xml");
+        FormEntryController fec = initFormEntry(fpi);
+        EvaluationContext evalCtx = fpi.getFormDef().getEvaluationContext();
+
+        ExprEvalUtils.assertEqualsXpathEval(
+                "Test that xforms-ready event triggered the form-level setvalue action",
+                "default value", "/data/selection", evalCtx);
+
+        Calendar birthday = Calendar.getInstance();
+        birthday.set(1993, 2, 26);
```
phillipm: linter is asking for `2` -> `Calendar.FEBRUARY`

core/src/test/resources/xform_tests/test_question_level_actions.xml
```diff
@@ -0,0 +1,61 @@
...
+        <model>
+            <instance>
+                <data
+                    name="question actions"
+                    xmlns="http://openrosa.org/test/actions"
+                    uiVersion="1"
+                    version="1">
+                    <text time=""/>
+                    <selection time=""/>
+                    <default/>
```
phillipm: is `<default/>` needed?

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -247,6 +237,68 @@ public void testRepeatInsertTriggering() throws Exception {
...
+            } else if (questionIndex == 2) {
+                fec.answerQuestion(new DateData(birthday.getTime()));
+            }
+
+            questionIndex++;
+        } while (fec.stepToNextEvent() != FormEntryController.EVENT_END_OF_FORM);
+
+        Object evalResult = ExprEvalUtils.xpathEval(evalCtx, "/data/text/@time");
+        assertTrue("Check that a timestamp was set for the text question",
+                evalResult.getClass().equals(Date.class));
```
phillipm: `evalResult.getClass().equals(Date.class)` -> `evalResult instanceof Date`

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
amstone326: Damn, thought I removed all of those, it did something weird when I moved that class. Thanks.

core/src/main/java/org/javarosa/core/model/actions/ActionController.java
amstone326: Hm yea, I generally like the clarity of overloading methods, but see the other side of the argument. Will probably leave for now.

core/src/main/java/org/javarosa/form/api/FormEntryCaption.java
amstone326: Ah thanks, stopped using it when I refactored my approach

phillipm:
Done with a pass; looks :moneybag: 
Left some code cleanup-related comments

core/src/main/java/org/javarosa/core/model/QuestionDef.java
amstone326: Yea, it's pretty abstracted away.. In `XFormPasrser.parseControlChildren()` if `actionHandlers` contains a handler for the element, it calls handle on the element, with its parent `QuestionDef`as the `parent` argument to `handle()`. If you follow that all the way down the chain of `parseAction()` --> `parseSetValueAction()` that `parent` object from the original call to `handle()` is passed to `parseSetValueAction()` as the `source` argument, and calls `source.registerEventListener(event, action)`.

amstone326: ..I can explain in person if that was hard to follow

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
amstone326: (2 actually corresponds to Calendar.MARCH, but changing)

core/src/test/resources/xform_tests/test_question_level_actions.xml
amstone326: Nope, leftover cruft from the original example Clayton gave me

amstone326:
@phillipm Thanks for review, made all minor changes. Pretty sure you're right about needing a migration :( Will do that and the serialization tests

core/src/main/java/org/javarosa/core/model/actions/Action.java
```diff
@@ -37,9 +46,7 @@ public Action(String name) {
      * @return TreeReference targeted by the action or null if the action
      * wasn't completed.
      */
-    public TreeReference processAction(FormDef model, TreeReference context) {
-        return null;
-    }
+    public abstract TreeReference processAction(FormDef model, TreeReference context);
```
phillipm: I tried to make `processAction` abstract before and it broke something in the J2ME build (https://github.com/dimagi/javarosa/pull/186). @ctsims  do you know what was going on?

amstone326:
@phillipm Added testing for `FormDef` serialization. This is being done by just running the test that I wrote for actions on a serialized-and-then-deserialized version of the `FormDef`. @ctsims said that this was the best way to do this sort of test, because defining equality of an object as complex as a `FormDef` isn't realistic, so is much easier to just test by basically saying "Can we still do all the same stuff with this object?". Clayton also said that explicit tests for `QuestionDef` serialization are unnecessary because that will be a major part of any `FormDef` serialization.

phillipm:
:+1: 
I'll let @ctsims do a pass too before merging since he has more context on this feature.

core/src/main/java/org/javarosa/core/model/actions/Action.java
```diff
@@ -9,14 +10,22 @@
...
-public class Action implements Externalizable {
+public abstract class Action implements Externalizable {
+
+    // Events that can trigger an action
     public static final String EVENT_XFORMS_READY = "xforms-ready";
     public static final String EVENT_XFORMS_REVALIDATE = "xforms-revalidate";
     public static final String EVENT_JR_INSERT = "jr-insert";
+    public static final String EVENT_QUESTION_VALUE_CHANGED = "xforms-value-changed";
+    private static final Vector<String> allEvents =
+            new Vector<String>(Arrays.asList(EVENT_JR_INSERT, EVENT_QUESTION_VALUE_CHANGED,
```
phillipm: `Arrays.asList` isn't J2ME compatible :tired_face:

amstone326: :disappointed: of course not

core/src/main/java/org/javarosa/core/model/actions/Action.java
ctsims: We used to need Action to be non-abstract because we were serializing it directly, if I recall correctly.

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -55,7 +56,8 @@
  *
  * @author Daniel Kayiwa, Drew Roos
  */
-public class FormDef implements IFormElement, Localizable, Persistable, IMetaData {
+public class FormDef extends ActionController
```
ctsims: It seems a bit unwieldy for me to require action triggering elements to extend the controller by their nature. the class already has to implement the processor, it seems like it would be cleaner to me to have an interface which exposed access to a controller object by composition. IE: The formdef has an "ActionController" object that handles serialization and the events/dispatch, and you call formDef.getActionController().registerEventListener() rather than making it "intrinsic" to the class itself.

amstone326: @ctsims The most recent commit makes the suggested change. I'm not sure whether or not I find it better. There are some trade-offs in terms of now having to repeat the same code in both `FormDef` and `QuestionDef`, but I do agree that having the original way was a bit heavy-handed for representing the true state of things. Up to you which seems better, or if the change I made can be improved upon still.

phillipm:
@amstone326 @ctsims Wanted to double check that a migration isn't needed: https://github.com/dimagi/javarosa/pull/243#issuecomment-181446079 
Also, @amstone326 moving `Action` will cause serialization issues if `Action` is ever externalizaed using a tag (for polymorphic lists/tables). Can you checkout 2.26 and verify that it wasn't being used with tags? If it is, then we need to move it back to its old file location.

amstone326:
@ctsims- @phillipm and I just chatted about this and *think* that this is OK because the only place where `Action` is externalized using a tag is in `ActionController`, which in turn is only externalized in `QuestionDef` and `FormDef`. My recollection with both of those classes is that if deserialization fails (which it will since I added the `ActionController` field to both of their `readExternal` methods), we just re-parse from the xform from scratch, meaning that `ActionController` won't get deserialized using tags anyway. Can you confirm that this all sounds correct?

PR #245, opened 2016-02-08
fix bug in deserialization of DateData

amstone326:
mis-ordered calls in `readExternal` cause a NPE when DateDatas are deserialized. This particular situation likely never actually comes up, for reasons Clayton explained and I did not really follow well enough to repeat here... But I came across it in writing a test, and seems good to fix either way.


phillipm:
I like how every method in this class calls `init()`...

PR #246, opened 2016-02-11
Linting CLI related code

phillipm:


core/src/main/java/org/javarosa/form/api/FormEntryController.java
```diff
@@ -255,9 +235,6 @@ public FormIndex getAdjacentIndex(FormIndex index, boolean forward, boolean expa
             //reset all step rules
             descend = true;
             relevant = true;
-            inForm = true;
```
phillipm: Set directly below

PR #248, opened 2016-02-12
Capture & replay form entry session

phillipm:
Capture form entry sessions for CLI and Android.

For CLI the following 2 form entry commands have been added:
 - `:entry-session` prints the current form entry session
 - `:replay the-session-string` replays the form entry session provided, which should be acquired by the `:entry-session` command

I will update the cli confluence docs when this gets merged


core/src/main/java/org/javarosa/form/api/FormEntrySession.java
```diff
@@ -0,0 +1,214 @@
...
+import org.javarosa.core.model.FormIndex;
+
+import java.io.Serializable;
+import java.util.Vector;
+
+/**
+ * Records form entry actions, associating form indexes with user (string)
+ * answers.  For simplicity's sake each form index appears only once in the
+ * action list. If a question answer is updated, the original form index entry
+ * is removed, and the updated entry is added to the end.
```
amstone326: This comment seems inconsistent with the code; when an answer is updated, you're placing the updated entry at the same index in the vector as the original answer, not at the end.

core/src/main/java/org/javarosa/form/api/FormEntrySession.java
```diff
@@ -0,0 +1,214 @@
...
+        public static FormEntryAction buildNullAction() {
+            return new FormEntryAction("", "", "");
+        }
+
+        @Override
+        public String toString() {
+            if (NEW_REPEAT.equals(action)) {
+                return "((" + formIndexString + ") (NEW_REPEAT))";
+            } else if (VALUE.equals(action)) {
+                // TODO PLM: escape 'value' field
```
amstone326: Is this TODO intentionally not getting done in this PR?

core/src/main/java/org/javarosa/form/api/FormEntrySession.java
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        public static FormEntryAction fromString(String entryActionString) {
+            String unwrappedEntryActionString =
+                    entryActionString.substring(1, entryActionString.length() - 1);
+            Vector<String> actionEntries = splitTopParens(unwrappedEntryActionString);
+            int entryCount = actionEntries.size();
+
+            if (entryCount != 2 && entryCount != 3) {
+                throw new RuntimeException();
```
amstone326: Is there anything else we can do here? Or just adding a message at least

core/src/main/java/org/javarosa/form/api/FormEntrySessionReplayer.java
```diff
@@ -0,0 +1,89 @@
...
+    public static void tryReplayingFormEntry(FormEntryController formEntryController,
+                                             FormEntrySession formEntrySession) {
+        FormEntrySessionReplayer replayer =
+                new FormEntrySessionReplayer(formEntryController, formEntrySession);
+        if (replayer.isRestoringFormSession()) {
+            replayer.replayForm();
+        }
+    }
+
+    private boolean isRestoringFormSession() {
```
amstone326: Nitpicking, but method name here seems weird. What you're checking is more like `hasFormSessionToReplay`, right?

amstone326: Ah I see, I hadn't seen the other usages of this method below. I guess ok to leave then

core/src/main/java/org/javarosa/form/api/FormEntrySessionReplayer.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    private void replayEvent(int event) {
+        if (event == FormEntryController.EVENT_QUESTION) {
+            replayQuestion();
+        } else if (event == FormEntryController.EVENT_PROMPT_NEW_REPEAT) {
+            if (formEntrySession.peekAction().isNewRepeatAction()) {
+                formEntryController.newRepeat();
+                while (formEntrySession.peekAction().isNewRepeatAction()) {
```
amstone326: Confused by what this while loop is doing. Why do you want to effectively ignore any subsequent 'new repeat' actions that come after an initial one

core/src/main/java/org/javarosa/form/api/FormEntrySessionReplayer.java
```diff
@@ -0,0 +1,89 @@
...
+
+        if (!questionIndex.toString().equals(action.formIndexString)) {
+            throw new ReplayError("Unable to replay form due to incorrect question index");
+        }
+
+        if (action.isSkipAction()) {
+            formEntrySession.popAction();
+            return;
+        }
+        while (questionIndex.toString().equals(formEntrySession.peekAction().formIndexString)) {
```
amstone326: I'm sure I'm not following this correctly, but this loop looks to me like it would just take you through all the way to the end of the form (so long as the form indices continue to match up). What am I missing?

amstone326:
Left a few small comments here and in the commcare-odk PR, but this looks awesome.

core/src/main/java/org/javarosa/form/api/FormEntrySessionReplayer.java
phillipm: This was an artifact of a bug from earlier on in this PR where each action would get added twice when filling out forms. I've fixed it, so I'll replace the `while` with an `if`. Sorry about the confusion.

core/src/main/java/org/javarosa/form/api/FormEntrySessionReplayer.java
phillipm: This `while` is here for the same reason as the `while` above. I've removed it too

core/src/main/java/org/javarosa/form/api/FormEntrySession.java
phillipm: Yeah, my hope for this was to work on it incrementally as it becomes more useful, since it is a side project of sorts. For instance, I don't think question lists are handled quite correctly yet, but I will fix that when I start needed to debug forms with question lists, or support does

core/src/main/java/org/javarosa/form/api/FormEntrySession.java
phillipm: I added a message. Don't really want to add comprehensive error handling at this point because I'm the only one using this. Currently when this `RuntimeException` is raised it crashes the app

phillipm:
@amstone326 thanks! Updated with your feedback.

core/src/main/java/org/javarosa/form/api/FormEntrySession.java
amstone326: @phillipm Hey, still think this comment is wrong. Good to merge once that's fixed

phillipm: @amstone326 Oops, forgot about this one, updated it

PR #250, opened 2016-02-19
constraint -> validation

wpride:
We call constraints validations in public facing text

See http://manage.dimagi.com/default.asp?217679


czue:
:+1: on :white_check_mark:

PR #251, opened 2016-02-19
Trigger lookaheads correctly in model iteration

phillipm:
Android-level 'form save on complete' re-fires all triggerables.  Disabling this, for performance reasons, exposes a bug were triggerables that fire on repeat inserts and have expressions that reference future (to be inserted) repeat entries never get re-fired due to over contextualization. This has been fixed by generalizing the triggerable context where path predicates are present for intersecting references in the triggerable expression.

For instance:
```
<bind 
    calculate="/data/myiterator/iterator[@id = current()/../target]/value" 
    nodeset="/data/myiterator/iterator/target_value"/>
```
This calculation triggerable should fire for all entries in `/data/myiterator/iterator` because `/data/myiterator/iterator[@id = current()/../target]/value` might reference an entry of `/data/myiterator/iterator` that doesn't yet exist. Making it fire for all entries means that when that entry exists (is created), the value will correctly get set.

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -593,12 +593,11 @@ public class FormDef implements IFormElement, Localizable, Persistable, IMetaDat
             triggerablesInOrder = false;
 
             for (TreeReference trigger : t.getTriggers()) {
-                t.updateStopContextualizingAt(trigger);
-                TreeReference plainTrigger = trigger.removePredicates();
-                if (!triggerIndex.containsKey(plainTrigger)) {
-                    triggerIndex.put(plainTrigger.clone(), new Vector<Triggerable>());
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -593,10 +593,12 @@ public Triggerable addTriggerable(Triggerable t) {
             triggerablesInOrder = false;
 
             for (TreeReference trigger : t.getTriggers()) {
-                if (!triggerIndex.containsKey(trigger)) {
-                    triggerIndex.put(trigger.clone(), new Vector<Triggerable>());
+                t.updateStopContextualizingAt(trigger);
+                TreeReference plainTrigger = trigger.removePredicates();
```
ctsims: TODO: Fold these two actions into the same method so people can't miss one,

PR #254, opened 2016-02-23
Fix form translate

wpride:
Fix to get Jenkins form translate build working


phillipm:
Is there a reason we aren't using gradle for this?

wpride:
No better reason than because I'm more comfortable with Ant and wanted to get these building today. If we think better to put on the backlog for Gradle I'm OK with that too

phillipm:
Is this for 2.26? If so we can add the changes to that branch and then not need to worry about reverting changes later when we migrate to gradle. Also, if it is for 2.26 you will need to make another PR including these changes for the 2.26 and update the 2.26.0 tag regardless, right?

If I'm off on this, I'm fine merging this.

wpride:
I've already generated the jar for 2.26 and its on staging in QA so that's good to go, since this won't impact any of the other builds I don't think putting on 2.26 and retagging and all is necessary

In terms of reverting, think that would mean just deleting these files and adding the Gradle ones so not too concerned

If we think this is too hacky/shitty to merge in I'm fine to close and try to get this working on Gradle for 2.27

PR #257, opened 2016-02-26
Prevent NPE on when setting TreeElement root to empty ref

phillipm:
Some refactors and a switch of order of ops to avoid an NPE

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/instance/TreeElement.java
+++ b/core/src/main/java/org/javarosa/core/model/instance/TreeElement.java
@@ -78,6 +78,16 @@ public class TreeElement implements Externalizable, AbstractTreeElement<TreeElem
 
     private String instanceName = null;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/javarosa/core/model/instance/TreeElement.java
```diff
@@ -146,18 +156,19 @@ public void setValue(IAnswerData value) {
...
     @Override
     public TreeElement getChild(String name, int multiplicity) {
-        if (this.children == null) {
+        if (children == null) {
             return null;
         }
 
-        if (name.equals(TreeReference.NAME_WILDCARD)) {
-            if (multiplicity == TreeReference.INDEX_TEMPLATE || this.children.size() < multiplicity + 1) {
+        if (TreeReference.NAME_WILDCARD.equals(name)) {
```
phillipm: Actual change, rest of PR is code readability

PR #261, opened 2016-03-01
General Linting

phillipm:


core/src/main/java/org/javarosa/xform/parse/XFormParser.java
```diff
@@ -1007,7 +1007,7 @@ protected QuestionDef parseControl(IFormElement parent, Element e, int controlTy
                             "of a <trigger> isn't allowed to have predicates.", e);
                 }
             } catch (RuntimeException el) {
-                System.out.println(this.getVagueLocation(e));
+                System.out.println(getVagueLocation(e));
```
orangejenny: Doesn't need to block this PR, but might be nice to write this to an actual log.

PR #264, opened 2016-03-04
Remove duplicate InvalidStructureException

phillipm:
fix for j2me build failures on jenkins where the wrong `InvalidStructureException` was being loaded for some reason. This PR makes it so the only version of `InvalidStructureException` that exists is https://github.com/dimagi/javarosa/blob/master/core/src/main/java/org/javarosa/xml/util/InvalidStructureException.java


orangejenny:
Looks reasonable.

PR #265, opened 2016-03-09
J2ME Build fix and Cert Redirect

ctsims:
Fixes the J2ME build (Serialization issues again)

Adds a platform redirect on J2ME where if an https request has certificate failures, it redirects the user to an optional URL that allows the user to create an exception on the device 

cross-request: https://github.com/dimagi/commcare/pull/267

### DIFF
--- a/j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
+++ b/j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
@@ -21,10 +21,13 @@ import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.Exception;
 import java.util.Date;
 
 import javax.microedition.io.Connector;
 import javax.microedition.io.HttpConnection;
+import javax.microedition.midlet.MIDlet;
+import javax.microedition.pki.CertificateException;
 
 import de.enough.polish.util.StreamUtil;
 
@@ -50,6 +53,8 @@ public class AuthenticatedHttpTransportMessage extends SimpleHttpTransportMessag
 
     HttpRequestProperties responseProperties;
     private SecurityFailureListener listener;
+
+    public static MIDlet CalloutResolver;
 
     private AuthenticatedHttpTransportMessage(String URL, HttpAuthenticator authenticator, SecurityFailureListener listener) {
         this.setCreated(new Date());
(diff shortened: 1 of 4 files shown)


j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
```diff
@@ -224,10 +230,13 @@ public void send() {
                 //handle the response.
                 handleResponse(connection);
             }
+        } catch (CertificateException e) {
+            //Certificates are now (March 2016) basically useless, since no one can issue a safe SHA1
+            //cert anymore. Direct the user to a browser to accept the authentication.
+            noteFailure(e);
+           	attemptBrowserCertificateAcceptCallout();
```
wpride: indentation

wpride: hm this indentation os wrong in my browser but not AS, nvm on comment

j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
```diff
@@ -241,6 +250,30 @@ public void send() {
...
+    
+    private void attemptBrowserCertificateAcceptCallout() {
+    	try {
+	    	if(CalloutResolver != null) {
+	    		//See if we have a location in our current context to do a cert accept
+	    		String fetchUrl = 
+	    				PropertyManager._().getSingularProperty(TransportPropertyRules.HTTP_CERTIFICATE_REQUEST_URL);
+	    		
+	    		if(fetchUrl == null) {
+	    			fetchUrl = URL;
```
wpride: comment and/or better name for "URL" would be useful - can't tell what this is

j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
```diff
@@ -241,6 +250,30 @@ public void send() {
             }
         }
     }
+    
+    private void attemptBrowserCertificateAcceptCallout() {
+    	try {
+	    	if(CalloutResolver != null) {
```
wpride: What situations will this be null/not null in? If its null, we can do nothing/should we notify the user?

j2me/communication/src/org/javarosa/services/transport/TransportPropertyRules.java
```diff
@@ -0,0 +1,114 @@
...
+		}
+		return false;
+	}
+
+	public boolean checkPropertyUserReadOnly(String propertyName) {
+		return readOnlyProperties.contains(propertyName);
+	}
+
+	/*
+	 * (non-Javadoc)
```
wpride: @phillipm I think usually asks that these "non-Javadoc" comments be removed

phillipm: haha, very true! That said, I don't care about code quality for J2ME because we rarely do dev for it, so I'm fine with this.

j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
ctsims: It's something that needs to get wired manually into the MIDlet (and would be optional for someone using the JR libraries), so shouldn't need to warn.

j2me/communication/src/org/javarosa/service/transport/securehttp/AuthenticatedHttpTransportMessage.java
ctsims: URL is an existing class member variable that is the URL for the Transport Message itself

j2me/communication/src/org/javarosa/services/transport/TransportPropertyRules.java
ctsims: As I mentioned this is copy/pasted from another file. I didn't really address any of the old stylistic problems in favor of not introducing new issues.

PR #266, opened 2016-03-10
Add back in meta data function used by TF

wpride:
Add back in function used by Touchforms removed in https://github.com/dimagi/javarosa/commit/cbea2ba72473c844b3bf08f59f32d760e91fa620

@phillipm I know Jython causes an annoying state where its very unclear what functions are being used, but can you refrain from removing more functions from javarosa/core and commcare/[backend, cases] for the time being? Hopefully in the near future (2-3 months) we'll be off Jython and the spring repository will make clear what functions are being used

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/instance/FormInstance.java
+++ b/core/src/main/java/org/javarosa/core/model/instance/FormInstance.java
@@ -270,6 +270,14 @@ public class FormInstance extends DataInstance<TreeElement> implements Persistab
         return new String[]{META_XMLNS, META_ID};
     }
 
+    public Hashtable getMetaData() {
+        Hashtable data = new Hashtable();
+        for (String key : getMetaDataFields()) {
+            data.put(key, getMetaData(key));
+        }
+        return data;
+    }
+
     @Override
     public Object getMetaData(String fieldName) {
         if (META_XMLNS.equals(fieldName)) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/javarosa/core/model/instance/FormInstance.java
```diff
@@ -270,6 +270,14 @@ public DataInstance initialize(InstanceInitializationFactory initializer, String
         return new String[]{META_XMLNS, META_ID};
     }
 
+    public Hashtable getMetaData() {
```
phillipm: Can you add a `@SuppressWarnings("unused")` annotation so that we know this is used by something outside of odk?

phillipm:
Yeah, sorry about that!

phillipm:
@wpride Do you know if any of the other methods I removed in https://github.com/dimagi/javarosa/commit/cbea2ba72473c844b3bf08f59f32d760e91fa620 are needed?

wpride:
@phillipm updated, thanks and np. 

I was looking through that and I don't believe so

ctsims:
Was this removed for coverage reasons, @phillipm ? If so would be good to put in a unit test or something so that we note that it is getting hit.

phillipm:
I just like deleting 'unused' code.
I've done a few dead code elimination passes of the entire codebase so I'm
pretty satisfied with the state of things. Hopefully shouldn't accidentally
be deleting used code anytime soon... Sorry that this caused issues!

PR #274, opened 2016-03-14
Don't allow ActionController to deserialize to null; prevents NPE crash

phillipm:
I've been coming against residual NPE crashes where form loading fails because the `FormDef`'s `ActionController` doesn't fail hard when deserialization fails, but rather to deserializes to `null`.

This PR makes form deserialization fail when the `ActionController` can't be deserialized. This forces the form to be re-read from xml and re-serialized.

To reproduce, load a form using an intent on master, then build https://github.com/dimagi/commcare-odk/pull/1128 and load the same form. The change in how `IntentCallout` is serialized somehow causes the `FormDef` deserialization to not load the `ActionController` correctly.

This change makes it so that if we change the serialization scheme of `ActionController`, or I guess other things as seen in the reproduction steps above, then we won't cause user level crashes when CommCare is updated on their devices and the load a form that had previously been loaded on older version of CommCare.

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1492,7 +1492,7 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
         extensions = (Vector)ExtUtil.read(dis, new ExtWrapListPoly(), pf);
 
         setEvaluationContext(new EvaluationContext(null));
-        actionController = (ActionController)ExtUtil.read(dis, new ExtWrapTagged(), pf);
+        actionController = (ActionController)ExtUtil.read(dis, ActionController.class, pf);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1492,7 +1492,7 @@ public void readExternal(DataInputStream dis, PrototypeFactory pf) throws IOExce
         extensions = (Vector)ExtUtil.read(dis, new ExtWrapListPoly(), pf);
 
         setEvaluationContext(new EvaluationContext(null));
-        actionController = (ActionController)ExtUtil.read(dis, new ExtWrapNullable(ActionController.class), pf);
+        actionController = (ActionController)ExtUtil.read(dis, new ExtWrapTagged(), pf);
```
phillipm: I'm pretty sure `ExtWrapTagged` is the right way to do this, but not totally sure...

amstone326:
@phillipm Hey, thanks for flagging this. I have no idea if `ExtWrapTagged` is the right alternative here, though- none of those `ExternalizableWrapper` classes have javadocs so it's kind of hard to tell. But if you tested and confirmed that it fixes the crash, then happy to merge.

PR #275, opened 2016-03-17
Make form replay more robust

amstone326:
Makes form replay robust to addition and deletion of questions, by changing 3 things:

1. Instead of matching widgets to actions using the FormIndex, use the actual path of the question (e.g. /data/name). This value works better as a unique identifier for a question/answer pairing because while app builders can change the question ID of a question, they do so much less often than they insert or delete questions. (And in the event that they have changed a question ID, all this will do is prevent the question from being repopulated; it won't break anything).

2. Don't assume that questions have to be replayed in the exact order that the set of `FormEntryAction`s is in. Instead, as you move through each question index in the form, just search the list of actions for an action with a matching reference.

3. If there is no action entry for the current index, don't throw an error, and instead just proceed (this allows for insertion of new questions).

### DIFF
--- a/core/src/main/java/org/javarosa/form/api/FormEntrySession.java
+++ b/core/src/main/java/org/javarosa/form/api/FormEntrySession.java
@@ -20,7 +20,9 @@ import java.util.Vector;
  * @author Phillip Mates (pmates@dimagi.com).
  */
 public class FormEntrySession implements FormEntrySessionRecorder, Externalizable {
+
     private Vector<FormEntryAction> actions = new Vector<FormEntryAction>();
+    private String sessionStopRef;
 
     /**
      * For Externalization
@@ -65,6 +67,18 @@ public class FormEntrySession implements FormEntrySessionRecorder, Externalizabl
         } else {
             return FormEntryAction.buildNullAction();
         }
+    }
+
+    /**
+     * @return the ref path for the last action in this form entry session (e.g. where the user
+     * stopped form entry)
+     */
+    public String getStopRef() {
+        return this.sessionStopRef;
+    }
+
+    private static String computeStopRef(Vector<FormEntryAction> actions) {
+        return actions.get(actions.size()-1).getQuestionRefString();
     }
 
     /**
@@ -115,6 +129,7 @@ public class FormEntrySession implements FormEntrySessionRecorder, Externalizabl
             formEntrySession.actions.addElement(FormEntryAction.fromString(actionString));
         }
 
+        formEntrySession.sessionStopRef = computeStopRef(formEntrySession.actions);
         return formEntrySession;
     }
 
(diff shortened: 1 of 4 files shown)


phillipm:
Good ideas! I assume that this works well with repeats by comparing the multiplicity in the repeat tree reference's string rep, like `"/data/patient[2]/name"`

amstone326:
@phillipm Yep, although that all conveniently happens under the hood, since the path returned by `FormIndex.getReference()` is already of that format.

amstone326:
@phillipm This is good to review. I tested it fairly thoroughly, including adding and removing questions before and after the stopping point, adding questions into question lists, and a bunch of messing around with repeats, all of which worked well. Would be great if you can see if you can come up with any failure modes I haven't thought of though.

amstone326:
ah ok have to fix the tests though

core/src/main/java/org/javarosa/form/api/FormEntrySession.java
```diff
@@ -47,34 +49,63 @@ private int removeDuplicateAction(String formIndexString) {
...
+
+    /**
+     * Remove and return the FormEntryAction corresponding to the given FormIndex, if there is
+     * one in this session
+     */
+    public FormEntryAction getAndRemoveActionForRef(TreeReference questionRef) {
+        for (int i = 0; i < actions.size(); i++) {
+            FormEntryAction action = actions.get(i);
+            if (action.getQuestionRefString().equals(questionRef.toString())) {
+                return actions.remove(i);
```
phillipm: `Vector.remove` isn't J2ME compat :)

core/src/main/java/org/javarosa/form/api/FormEntrySessionReplayer.java
```diff
@@ -26,47 +27,58 @@ public static void tryReplayingFormEntry(FormEntryController formEntryController
...
         return formEntrySession != null && formEntrySession.size() > 0;
     }
 
+    /**
+     * TODO AMS: If the question corresponding to the stopping ref has been removed, this will
+     * never return true and replay will take the user all the way to the end of the form
+     */
+    private boolean reachedEndOfReplay(String lastQuestionRefReplayed) {
+        return lastQuestionRefReplayed.equals(formEntrySession.getStopRef());
+    }
```
phillipm: You could store the number of answers in the form entry session and then count how many questions got answered during replaying.

amstone326: Doesn't that essentially come down to the same thing as checking if the vector of actions is empty?  If one of the questions that was originally answered in the form entry session has been subsequently deleted, the count would never reach zero. This is at least a little more robust because it only fails if 1 specific question was deleted.

amstone326:
Nice, thanks for the j2me fix and quick merge!

PR #278, opened 2016-03-23
Notify user of self-indexing case in restore payload

phillipm:


core/src/main/java/org/javarosa/xml/util/ActionableInvalidStructureException.java
```diff
@@ -0,0 +1,32 @@
...
+    public ActionableInvalidStructureException(String localizationKey,
+                                                String[] localizationParameters,
+                                                String message) {
+        super(message);
+        this.localizationKey = localizationKey;
+        this.localizationParameters = localizationParameters;
+    }
+
+    @Override
+    public String getLocalizedMessage() {
```
phillipm: We should try to localize error messages when appropriate by overriding `Throwable.getLocalizedMessage`

PR #280, opened 2016-03-27
More progress towards making TreeReference immutable

phillipm:


core/src/main/java/org/javarosa/core/model/instance/TreeReference.java
```diff
@@ -128,13 +127,11 @@ public static TreeReference buildRefFromTreeElement(AbstractTreeElement elem) {
             TreeReference step;
 
             if (elem.getName() != null) {
-                step = new TreeReference(elem.getInstanceName(), REF_ABSOLUTE, CONTEXT_ABSOLUTE);
+                step = new TreeReference(elem.getInstanceName(), 0, CONTEXT_INHERITED);
```
amstone326: confused by this change, what's going on here?

phillipm: I think this was a typo in this commit that was fixed later. Basically I mimicking some of the setup that `selfRef` does and then letting the `setupContextTypeFromInstanceName`, which is called in the constructor, do the rest of the work

PR #283, opened 2016-04-01
General Linting

phillipm:


core/src/main/java/org/javarosa/core/util/externalizable/MD5Hasher.java
```diff
@@ -1,27 +0,0 @@
...
- * Default class hasher implementation.
- *
- * Based on in-house JavaRosa MD5'er
- *
- * Not super efficient on platforms where faster hashing
- * is available through other libraries
- *
- * @author ctsims
- */
-public class MD5Hasher extends Hasher {
```
phillipm: @wpride is this used by Touchforms? I don't think it is used anywhere else

amstone326: bump @wpride

wpride: @phillipm @amstone326 nope, this is the old one that was deprecated by the classnamehasher. Should have been removed then, my b

PR #284, opened 2016-04-05
Fix binary op expression equality logic

amstone326:
Fix for binary operation expression equality bug introduced in  https://github.com/dimagi/javarosa/pull/168 where two different operation classes, like equality and comparative, could be equal if they had the same sub-expressions and operator number encoding.

Basically, 'a < b' and 'a = b' were being treated as the same triggerable condition.

Addresses this ticket http://manage.dimagi.com/default.asp?222850

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -750,7 +750,7 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
                 // when we grabbed all triggerables that target children of the
                 // repeat entry (via initTriggerablesRootedBy). Hence skip them
                 if (!isRepeatEntryInit && t.isCascadingToChildren()) {
-                    //updatedNodes = findCascadeReferences(target, updatedNodes);
+                    updatedNodes = findCascadeReferences(target, updatedNodes);
                 }
 
                 addTriggerablesTargetingNodes(updatedNodes, destination);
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -750,7 +750,7 @@ private void fillTriggeredElements(Triggerable t,
                 // when we grabbed all triggerables that target children of the
                 // repeat entry (via initTriggerablesRootedBy). Hence skip them
                 if (!isRepeatEntryInit && t.isCascadingToChildren()) {
-                    updatedNodes = findCascadeReferences(target, updatedNodes);
+                    //updatedNodes = findCascadeReferences(target, updatedNodes);
```
amstone326: @phillipm Had me try commenting out a few different lines in his optimizations to see if that fixed it, but none of these changes had an impact.

core/src/test/resources/xform_tests/test_display_conditions_regression.xml
```diff
@@ -0,0 +1,46 @@
...
+                    <threshold>0</threshold>
+                    <labels>
+                        <label_should_show />
+                        <label_should_not_show />
+                    </labels>
+                </data>
+            </instance>
+
+            <bind nodeset="/data/labels/label_should_show" relevant="/data/threshold = 0" />
+            <bind nodeset="/data/labels/label_should_not_show" relevant="/data/threshold &lt; 0" />
```
amstone326: @ctsims The test will fail with the relevant expression here, but will pass if you change it to `relevant="/data/threshold &lt; -1"`

core/src/main/java/org/javarosa/core/model/FormDef.java
phillipm: Sorry, didn't mean for you to comment this out... probably messes something up if you do

phillipm:
I have a vague suspicion that https://github.com/dimagi/javarosa/pull/175 introduced the underlying issue.

core/src/main/java/org/javarosa/xpath/expr/XPathBinaryOpExpr.java
```diff
@@ -31,13 +31,10 @@ public String toString(String op) {
...
 
     @Override
-    public boolean equals(Object o) {
-        if (o instanceof XPathBinaryOpExpr) {
-            XPathBinaryOpExpr x = (XPathBinaryOpExpr)o;
-            return op == x.op && a.equals(x.a) && b.equals(x.b);
-        } else {
-            return false;
-        }
+    public abstract boolean equals(Object o);
```
ctsims: I don't think this is necessary, and I think it will break J2ME (or maybe it's the opposite.... It's one of the two, we will want to test a form which triggers an equality check on one of the implementing classes with the 2.27 build after it is pulled)

PR #286, opened 2016-04-06
Allow TreeReferenceLevel to be serialized if it has predicates

phillipm:
So far just failing test case


phillipm:
Caused by incorrectly setting up the prototype factory or not passing the prototype factory into a read method. I'll follow up with a PR that fixes all these instances

PR #290, opened 2016-04-14
Fix for IText Calculations not properly initializing with the app locale

ctsims:
Fix for bugs where the itext calculation in xpath was always using the form's default locale, even if the app locale is different.

Needed for ICDS

Fogbugz:
http://manage.dimagi.com/?221785

cross-request: https://github.com/dimagi/commcare-odk/pull/1200

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -1501,7 +1501,20 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
      * @param newInstance true if the form is to be used for a new entry interaction,
      *                    false if it is using an existing IDataModel
      */
+
     public void initialize(boolean newInstance, InstanceInitializationFactory factory) {
+        initialize(newInstance, factory, null);
+    }
+
+    /**
+     * meant to be called after deserialization and initialization of handlers
+     *
+     * @param newInstance true if the form is to be used for a new entry interaction,
+     *                    false if it is using an existing IDataModel
+     * @param locale The default locale in the current environment, if provided. Can be null
+     *               to rely on the form's internal default.
+     */
+    public void initialize(boolean newInstance, InstanceInitializationFactory factory, String locale) {
         for (Enumeration en = formInstances.keys(); en.hasMoreElements(); ) {
             String instanceId = (String)en.nextElement();
             DataInstance instance = formInstances.get(instanceId);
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -431,14 +432,62 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
         }
     }
 
+
+
+    /**
+     * Regression: IText function in xpath was not properly using the current locale instead of the
+     * backup
```
phillipm: Can you change 'backup' to 'default' here?

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -431,14 +432,62 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+     */
+    @Test
+    public void testITextXPathFunction() throws XPathSyntaxException {
+        FormParseInit fpi = new FormParseInit("/xform_tests/itext_function.xml");
+        FormEntryController fec =  initFormEntry(fpi, "new");
```
phillipm: can you add a comment that says we are using the 'new' locale instead of 'old', which is the form's default?

core/src/test/resources/xform_tests/itext_function.xml
```diff
@@ -0,0 +1,48 @@
+<h:html xmlns:h="http://www.w3.org/1999/xhtml"
+        xmlns="http://www.w3.org/2002/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+        xmlns:jr="http://openrosa.org/javarosa">
+    <h:head>
+        <h:title>Constraints for simple and select answers</h:title>
```
phillipm: Might be worth updating, sometimes it is hard to get context from test forms and the title helps

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1512,8 +1525,14 @@ public void initialize(boolean newInstance, InstanceInitializationFactory factor
...
-        if (getLocalizer() != null && getLocalizer().getLocale() == null) {
-            getLocalizer().setToDefault();
+        if (getLocalizer() != null) {
+            if(locale == null || !getLocalizer().hasLocale(locale)) {
+                if(getLocalizer().getLocale() == null) {
+                    getLocalizer().setToDefault();
+                }
+            } else {
+                getLocalizer().setLocale(locale);
+            }
```
phillipm: Can you lift this code into a helper, it is pretty contained

phillipm:
Looks solid, I'll merge after you address my minor readability requests

PR #292, opened 2016-05-02
Fix for bad error wrapping

ctsims:
Significantly better handling of encoding errors when serializing documents, lets errors bubble up semantically.

### DIFF
--- a/core/src/main/java/org/javarosa/xform/util/XFormSerializer.java
+++ b/core/src/main/java/org/javarosa/xform/util/XFormSerializer.java
@@ -6,6 +6,7 @@ import org.kxml2.kdom.Element;
 
 import java.io.ByteArrayOutputStream;
 import java.io.DataOutputStream;
+import java.io.IOException;
 import java.io.OutputStreamWriter;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/javarosa/xform/util/XFormSerializer.java
```diff
@@ -34,18 +35,39 @@ public static String elementToString(Element e) {
...
+
+    /**
+     * Formats an XML document into a UTF-8 (no BOM) compatible format
+     *
+     * @return The raw bytes of the utf-8 encoded doc
+     * @throws IOException If there is an issue transferring the bytes to a byte stream.
+     * @throws UnsupportedEncodingException If the document contains values that are not UTF-8
+     * encoded.
+     */
+    public static byte[] getUtfBytesFromDocument(Document doc) throws IOException {
```
phillipm: Don't you want to change https://github.com/dimagi/javarosa/blob/master/core/src/main/java/org/javarosa/model/xform/XFormSerializingVisitor.java#L81 to use this new method that propagates the exception?

PR #294, opened 2016-05-06
Make validator write unicode output

phillipm:
Make the form_translate jar output UTF-8 when possible. This will make error messages with unicode more helpful on HQ.

Instead of seeing `?????` in the image below, we will see real unicode
2016-05-06-101354_2400x1920_scrot

Changes `form_translate.jar validate` output from:
```
{"problems":[],"fatal_error":"Invalid XPath in value set action declaration: '??????'\n    Problem found at nodeset: \/html\/head\/model\/setvalue\n    With element <setvalue event=\"xforms-ready\" ref=\"\/data\/emoji_list\/question3\" value=\"??????\"\/>\n","fatal_error_expected":true,"validated":false}
```
to
```
{"problems":[],"fatal_error":"Invalid XPath in value set action declaration: 'हिन्दी'\n    Problem found at nodeset: \/html\/head\/model\/setvalue\n    With element <setvalue event=\"xforms-ready\" ref=\"\/data\/emoji_list\/question3\" value=\"हिन्दी\"\/>\n","fatal_error_expected":true,"validated":false}
```

Fix for this ticket: http://manage.dimagi.com/default.asp?113905

@millerdev: might find it nice to see a 2 year old ticket getting closed out

### DIFF
--- a/util/schema-gen/src/org/javarosa/xform/schema/Harness.java
+++ b/util/schema-gen/src/org/javarosa/xform/schema/Harness.java
@@ -77,7 +77,7 @@ public class Harness {
             try {
                 xfp.start(leftOverArgs[1]);
             } catch (FileNotFoundException e) {
-                System.out.println("File not found at " + args[0]+ "!!!!");
+                System.out.println("File not found at " + args[0] + "!!!!");
             }
 
         } else {
@@ -208,10 +208,16 @@ public class Harness {
                 System.exit(1);
             }
         }
-        PrintStream responseStream = System.out;
+        PrintStream defaultOutStream = System.out;
+        PrintStream responseStream;
         // Redirect output to syserr because sysout is being used for the
         // response, and must be kept clean.
-        System.setOut(System.err);
+        try {
+            responseStream = new PrintStream(System.err, false, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            responseStream = System.err;
+        }
+        System.setOut(responseStream);
 
         InputStreamReader isr;
         try {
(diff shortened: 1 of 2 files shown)


util/schema-gen/src/org/javarosa/xform/schema/Harness.java
```diff
@@ -376,8 +382,14 @@ private static void processSchema(FormDef form) {
...
-        PrintStream responseStream = System.out;
-        System.setOut(System.err);
+        PrintStream defaultOutStream = System.out;
+        PrintStream responseStream;
+        try {
+            responseStream = new PrintStream(System.err, false, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            responseStream = System.err;
+        }
+        System.setOut(responseStream);
```
phillipm: Not part of the fix, figured I'd include it none the less

util/schema-gen/src/org/javarosa/xform/schema/Harness.java
```diff
@@ -208,10 +208,16 @@ private static void validateForm(String[] args) {
...
+        PrintStream responseStream;
         // Redirect output to syserr because sysout is being used for the
         // response, and must be kept clean.
-        System.setOut(System.err);
+        try {
+            responseStream = new PrintStream(System.err, false, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            responseStream = System.err;
+        }
+        System.setOut(responseStream);
```
phillipm: The fix

millerdev:
yay!

util/schema-gen/src/org/javarosa/xform/schema/Harness.java
phillipm: Weird, looks like there is some sort of strange github bug where in `File changes` view, the comment above (attached to line 392) shows in the wrong place...

PR #296, opened 2016-05-15
Tell user when `replace` uses bad regexp

phillipm:
Show the user an error when they use a bad regexp in a `replace` xpath expression, like 
 `replace("abbc", "a(.*c", "$1")`

screen


orangejenny:
nice

snopoke:
Realise this isn't part of the code in this PR but is it necessary to have `XPath evaluation` in the text? I doubt most of our users know what that means and it just adds technical jargon that doesn't really add much value.
```
The problem was located in the calculate expression for the '/data/last_watched_years' question.
The regular expression 'a(.*c' is invalid.
```

PR #299, opened 2016-05-25
Better missing itemset error message

phillipm:


core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -1333,13 +1333,14 @@ public void populateDynamicChoices(ItemsetBinding itemset, TreeReference curQRef
                 new EvaluationContext(exprEvalContext, itemset.contextRef.contextualize(curQRef)));
 
         if (matches == null) {
-            throw new XPathException("Could not find references depended on by " + itemset.nodesetRef.getInstanceName());
+            throw new XPathException("No items found at '" + itemset.nodesetRef +
```
amstone326: If the issue that we expect to cause this is users not having access to the items in the table, I'm not sure that this error message quite captures that. Are you intentionally not mentioning that cause specifically?

phillipm: This error can pop up from non-lookup table select questions, so I wanted to leave it sorta vague. Is there a way to make `Make sure the 'someInstance' lookup table is available` clearer?

amstone326: I would maybe just extend it to say `Make sure the 'someInstance' lookup table is available and that its contents are accessible to the current user`, or something along those lines?

amstone326: @phillipm Waiting on this change to merge (or let me know if there's a reason you don't want to make the change)

phillipm: Updated to include that message, but only if the itemset references a lookup table, that is, an external data instance.

PR #304, opened 2016-06-14
Use (much faster) built-in String.split() method when possible

amstone326:
As part of performance improvements for ICDS Daily Feeding Form: Using built in String.split() method instead of DateUtils.split() improves form load time with a case load of 400 from 1 min 40 seconds to 27 seconds


phillipm:
🚀 🚄 🚀!

PR #309, opened 2016-06-16
Repeat delete fix

wpride:
Fix for: http://manage.dimagi.com/default.asp?229136

Appears that this has been broken for a while, but is (obviously) a very rare situation. Essentially, when an element was deleted we were grabbing all of its parent's children then iterating through these and reducing the multiplicity if they exceeded a certain threshold. This works fine when a repeat is within a group or there are no other repeats, but when (as in this case) the repeat's parent element has other repeat elements as children AND these children's multiplicities exceed the threshold, then we incorrectly reduce the multiplicities. The solution is to check that the deleted node and the child nodes are the "same" in every sense except multiplicity. 

This fix is admittedly very inelegant, but since the needed functionality is basically the equals function minus the multiplicity check I couldn't think of a prettier way to do it.

This should get into 2.28 so we can release the new TF jars.

### DIFF
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -360,9 +360,24 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
         TreeReference parentRef = deleteRef.getParentRef();
         TreeElement parentElement = mainInstance.resolveReference(parentRef);
 
-        int childMult = deleteElement.getMult();
         parentElement.removeChild(deleteElement);
 
+        reduceTreeSiblingMultiplicities(parentElement, deleteElement);
+
+        this.getMainInstance().cleanCache();
+
+        triggerTriggerables(deleteRef);
+        return newIndex;
+    }
+
+    /**
+     * When a repeat is deleted, we need to reduce the multiplicities of its siblings that were higher than it
+     * by one.
+     * @param parentElement the parent of the deleted element
+     * @param deleteElement the deleted element
+     */
+    private void reduceTreeSiblingMultiplicities(TreeElement parentElement, TreeElement deleteElement){
+        int childMult = deleteElement.getMult();
         // update multiplicities of other child nodes
         for (int i = 0; i < parentElement.getNumChildren(); i++) {
             TreeElement child = parentElement.getChildAt(i);
@@ -372,11 +387,6 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
                 child.setMult(child.getMult() - 1);
             }
         }
-
-        this.getMainInstance().cleanCache();
-
-        triggerTriggerables(deleteRef);
-        return newIndex;
     }
 
     public void createNewRepeat(FormIndex index) throws InvalidReferenceException {
--- a/core/src/main/java/org/javarosa/core/model/FormDef.java
+++ b/core/src/main/java/org/javarosa/core/model/FormDef.java
@@ -366,7 +366,9 @@ public class FormDef implements IFormElement, Persistable, IMetaData,
         // update multiplicities of other child nodes
         for (int i = 0; i < parentElement.getNumChildren(); i++) {
             TreeElement child = parentElement.getChildAt(i);
-            if (child.getMult() > childMult) {
+            // We also need to check that this element matches the deleted element (besides multiplicity)
+            // in the case where the deleted repeat's parent isn't a subgroup
+            if (child.doFieldsMatch(deleteElement) && child.getMult() > childMult) {
                 child.setMult(child.getMult() - 1);
             }
         }
--- a/core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
+++ b/core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
@@ -477,19 +477,19 @@ public class FormDefTest {
         TreeElement root = fpi.getFormDef().getInstance().getRoot();
 
         // confirm both groups have two iterations and second iteration is set
-        assert root.getChildMultiplicity("question4") == 2;
-        assert root.getChild("question4", 1) != null;
-        assert root.getChildMultiplicity("question1") == 2;
-        assert root.getChild("question1", 1) != null;
+        assertTrue(root.getChildMultiplicity("question4") == 2);
+        assertTrue(root.getChild("question4", 1) != null);
+        assertTrue(root.getChildMultiplicity("question1") == 20;
+        assertTrue(root.getChild("question1", 1) != null);
 
         fec.deleteRepeat(0);
 
         // Confirm that the deleted repeat is gone and its sibling's multiplicity reduced
-        assert root.getChildMultiplicity("question4") == 1;
-        assert root.getChild("question4", 1) == null;
+        assertTrue(root.getChildMultiplicity("question4") == 1);
+        assertTrue(root.getChild("question4", 1) == null);
         // Confirm that the other repeat is unchanged
-        assert root.getChildMultiplicity("question1") == 2;
-        assert root.getChild("question1", 1) != null;
+        assertTrue(root.getChildMultiplicity("question1") == 2);
+        assertTrue(root.getChild("question1", 1) != null);
     }
 
     /**
--- a/core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
+++ b/core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
@@ -31,9 +31,9 @@ import java.util.Calendar;
 import java.util.Date;
 import java.util.Vector;
 
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 import org.javarosa.xform.parse.XFormParser;
+
+import static org.junit.Assert.*;
 
 /**
  * @author Phillip Mates (pmates@dimagi.com)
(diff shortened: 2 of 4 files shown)


ctsims:
Will: Can you add a test for this? The test runner should be capable of replicating the original condition, right?

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -366,7 +366,9 @@ public FormIndex deleteRepeat(FormIndex index) {
         // update multiplicities of other child nodes
         for (int i = 0; i < parentElement.getNumChildren(); i++) {
             TreeElement child = parentElement.getChildAt(i);
-            if (child.getMult() > childMult) {
+            // We also need to check that this element matches the deleted element (besides multiplicity)
+            // in the case where the deleted repeat's parent isn't a subgroup
+            if (child.doFieldsMatch(deleteElement) && child.getMult() > childMult) {
```
phillipm: Took me a bit to understand what `child.doFieldsMatch(deleteElement)` is filtering in/out. Might help to mention repeat entry 'siblings'. That is this clause selects siblings of the deleted repeat entry and ignores other child of the parent element.

Maybe just lifting this whole for-loop into a helper called something like `reduceMultiplicityOfSiblingRepeatEntries` (or something shorter 😄 )

amstone326: Ah, agreed that that would help clarify a lot, I was a bit confused as well.

wpride:
New test fails when fix is not included

core/src/main/java/org/javarosa/core/model/FormDef.java
```diff
@@ -360,21 +360,33 @@ public FormIndex deleteRepeat(FormIndex index) {
...
+        return newIndex;
+    }
+
+    /**
+     * When a repeat is deleted, we need to reduce the multiplicities of its siblings that were higher than it
+     * by one.
+     * @param parentElement the parent of the deleted element
+     * @param deleteElement the deleted element
+     */
+    private void reduceTreeSiblingMultiplicities(TreeElement parentElement, TreeElement deleteElement){
```
phillipm: 💯

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -492,4 +532,21 @@ private static FormEntryController initFormEntry(FormParseInit fpi, String local
         fec.jumpToIndex(FormIndex.createBeginningOfFormIndex());
         return fec;
     }
+
+    String readFile(String fileName) throws IOException {
```
phillipm: Is this method used?

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +438,41 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        fec.answerQuestion(new StringData("Repeat 3/2"));
+        fec.stepToNextEvent();
+        fec.newRepeat();
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 4/1"));
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 4/2"));
+        fec.stepToPreviousEvent();
+        fec.stepToPreviousEvent();
+        fec.deleteRepeat(0);
```
phillipm: What repeat are we deleting here? The first one? or 3/2?
Can you add a check someone to ensure we don't decrement the multiplicity of entries from other repeats? or does it just crash if that code doesn't work properly? If so would still prefer an explicit check, but a comment might also suffice.

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,49 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+
+        fec.deleteRepeat(0);
+        TreeElement root = fpi.getFormDef().getInstance().getRoot();
+
+        // Confirm that the deleted repeat is gone and its sibling's multiplicity reduced
+        assert root.getChildMultiplicity("question4") == 1;
+        assert root.getChild("question1", 4) == null;
+        // Confirm that the other repeat is unchanged
+        assert root.getChildMultiplicity("question1") == 2;
+        assert root.getChild("question1", 1) != null;
```
phillipm: Can you add similar checks before `deleteRepeat` so we can more easily see what has changed?

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,49 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 1/2"));
+        fec.stepToNextEvent();
+        fec.newRepeat();
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 2/1"));
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 2/2"));
+        fec.stepToNextEvent();
+        fec.stepToNextEvent();
```
phillipm: Can you add a line-break here to show that we're going from `question1` to `question4`?

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,49 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        fec.newRepeat();
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 1/1"));
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 1/2"));
+        fec.stepToNextEvent();
+        fec.newRepeat();
+        fec.stepToNextEvent();
+        fec.answerQuestion(new StringData("Repeat 2/1"));
```
phillipm: Instead of `1/1` and `2/1` could you say something like `1st repeat, 1st entry` and `1st repeat, 2nd entry`

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,49 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        fec.answerQuestion(new StringData("Repeat 4/2"));
+        fec.stepToPreviousEvent();
+        fec.stepToPreviousEvent();
+
+        fec.deleteRepeat(0);
+        TreeElement root = fpi.getFormDef().getInstance().getRoot();
+
+        // Confirm that the deleted repeat is gone and its sibling's multiplicity reduced
+        assert root.getChildMultiplicity("question4") == 1;
+        assert root.getChild("question1", 4) == null;
```
phillipm: `assert` -> `assertTrue`

wpride: Why the preference for that?

phillipm: Mostly consistency: `assert` is Java syntax while `assertTrue` is a JUnit method. In tests we use `assertEquals`, `assertFalse`, etc. so it is nice to use `assertTrue` here . Also always for an optional message argument that describes why a test fails when it does.

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,57 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        assertTrue(root.getChild("question1", 1) != null);
+
+        fec.deleteRepeat(0);
+
+        // Confirm that the deleted repeat is gone and its sibling's multiplicity reduced
+        assertTrue(root.getChildMultiplicity("question4") == 1);
+        assertTrue(root.getChild("question4", 1) == null);
+        // Confirm that the other repeat is unchanged
+        assertTrue(root.getChildMultiplicity("question1") == 2);
+        assertTrue(root.getChild("question1", 1) != null);
```
phillipm: Last nitpick: instead of doing `assertTrue(a != b)` and `assertTrue(a == b) can you do `assertNotEquals(a, b)` and `assertEquals(a, b)`

When `b` is `null`: `assertNotNull(a)`

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,57 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        fec.answerQuestion(new StringData("Second repeat, second iteration: question6"));
+        fec.stepToPreviousEvent();
+        fec.stepToPreviousEvent();
+
+        TreeElement root = fpi.getFormDef().getInstance().getRoot();
+
+        // confirm both groups have two iterations and second iteration is set
+        assertTrue(root.getChildMultiplicity("question4") == 2);
+        assertTrue(root.getChild("question4", 1) != null);
+        assertTrue(root.getChildMultiplicity("question1") == 20;
```
phillipm: `20` -> `2`?

core/src/test/java/org/javarosa/core/model/test/FormDefTest.java
```diff
@@ -432,7 +440,57 @@ public void testSimilarBindConditionsAreDistinguished() throws Exception {
...
+        assertTrue(root.getChild("question1", 1) != null);
```
phillipm: Thanks for adding these pre-conditions, super helpful!

PR #313, opened 2016-07-01
Return result of union, not in-going set

phillipm:
Follow-up to https://github.com/dimagi/commcare-android/pull/1307 where I changed the behavior to not modify the vector arguments.

http://manage.dimagi.com/default.asp?231366


phillipm:
I'll follow up with a PR targeting `master` that renames `union` to `intersect` because the `union` code in fact performs a set intersect...

PR #315, opened 2016-07-06
Fix for J2ME Calculated Session Data

ctsims:
Replication of 
https://github.com/dimagi/commcare-core/pull/325

from 2.28


amstone326:
@ctsims I think we've started not creating 2 PRs for a fix that's going into the current release, and instead relying on the fact that we'll do a backmerge of (for instance) 2.28 into master

amstone326:
Ah nevermind sorry, just saw the convo on the other PR, makes sense

